Issue #2, opened 2019-08-21
Some CommonMark spec examples do not roundtrip


dselman:
The test is currently skipped here:
https://github.com/accordproject/markdown-transform/blob/master/src/CommonMark.test.js#L149

RounakKumarAgarwal:
I’d like to work on this issue. Please let me know if that’s okay.

PR #26, opened 2019-09-25
Refactor for API Consistency

dselman:

# Issue #30

Lots of API changes to make the naming more consistent across the various transformation layers. 
### Changes
- README.md updates
- Added a PlantUML architecture diagram
- API renaming for consistency
- Added lots of tests and fixed issues found

### Flags
- There are still some roundtrip test failures, being tracked under issue #2 
- Variable values are quoted: #34


packages/markdown-slate/package.json
```diff
@@ -76,7 +76,7 @@
     "webpack-cli": "^3.3.5"
   },
   "dependencies": {
-    "@accordproject/markdown-common": "0.3.1",
+    "@accordproject/markdown-cicero": "^0.3.1",
```
jeromesimeon: I think for the lerna packages, we have used exact versions in mutual dependencies (at least in Ergo and Cicero). I would probably keep it that way.

dselman: Fixed

PR #49, opened 2019-10-03
fix(slate): move text value from data to nodes

irmerk:
# No issue

### Changes
Moving the `text` value from `data` to `nodes`

### Flags
- Unable to resolve these two failing tests:
  - https://github.com/accordproject/markdown-transform/blob/3878ac6253a92f9c50713c9ba8cbe42b04273f7a/packages/markdown-cli/test/cli.js#L87
  - https://github.com/accordproject/markdown-transform/blob/3878ac6253a92f9c50713c9ba8cbe42b04273f7a/packages/markdown-cli/test/cli.js#L55
- Resolved some failing tests (`packages/markdown-slate/src/SlateTransformer.test.js`) by adding `'marks': [],`, which doesn't seem like it was necessary. This lead to the remaining failing tests, which a text diff checker finds that the error is the included `'marks': [],`

### DIFF
--- a/packages/markdown-cli/test/data/acceptance-slate.json
+++ b/packages/markdown-cli/test/data/acceptance-slate.json
@@ -85,12 +85,12 @@
                         "object":"inline",
                         "type":"variable",
                         "data":{
-                           "id":"shipper",
-                           "value":"%22Party%20A%22"
+                           "id":"shipper"
                         },
-                        "nodes":[
-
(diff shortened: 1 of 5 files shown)


jeromesimeon:
Just to not forget: I think this code misses the reverse transformation Slate -> CiceroMark which is likely why roundtrip tests are not passing.

PR #60, opened 2019-10-08
docs(README): add slack and build badge

sumitkharche:
### Changes
- added slack badge
- added build badge

### Related Issues
- accordproject/techdocs/issues/82
- accordproject/techdocs/issues/78


sumitkharche:
> Could use the slack badge implemented here: https://github.com/accordproject/techdocs

Updated the slack badge as per above link and also added build badge

PR #69, opened 2019-10-12
Add bootstrap

dlemstra:
### Changes
This PR adds `lerna bootstrap` as a postinstall step to make sure the packages are setup correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,6 @@ before_install: |
   export PATH="./$HOME/node_modules/.bin:$PATH"
   npm install -g @alrra/travis-scripts
   date
-install: |
-  set -e
-  date
-  npm install -g lerna@^3.15.0
-  lerna bootstrap 2>&1
-  date
 script: npm run test 2>&1
 deploy:
     provider: script
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,834 +1,14 @@
 {
 	"name": "markdown-transform",
-	"version": "0.4.6",
+	"version": "0.4.8",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"@accordproject/concerto-core": {
-			"version": "0.81.1",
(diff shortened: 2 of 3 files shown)


jeromesimeon:
@dlemstra We don't have an issue with this that I know of. Can you explain what the problem you are trying to solve is? If there is something wrong with CI, please open a new issue first. Thanks.

dlemstra:
This is not a CI issue. I was trying to see I could fix one of the issues and I noticed that `lerna bootstrap` was not automatically executed when an `npm install` was run. This PR fixes that. The CI changes were made because running lerna bootstrap is now done automatically in npm install.

p.s. I hope you don't close every PR without even giving the contributer time to respond.

jeromesimeon:
@dlemstra I only closed because I could not understand what this was about, as explained. It sounds like what you're talking about could be addressed by updating the documentation to clearly indicate that this is a lerna repo and to call `lerna bootstrap` not `npm install`. I believe we do that for other lerna repos we maintain: https://github.com/accordproject/cicero/blob/master/DEVELOPERS.md#building-cicero

jeromesimeon:
@dlemstra I think we want to call `lerna bootstrap` instead of `npm install`.

dlemstra:
Calling `lerna bootstrap` would require you to have lerna installed globaly with this setup you can use `npm install` that runs `lerna` after the packages are installed. But that is probably just a matter of preference.

jeromesimeon:
> Calling `lerna bootstrap` would require you to have lerna installed globaly with this setup you can use `npm install` that runs `lerna` after the packages are installed. But that is probably just a matter of preference.

Fair enough. I personally like the idea of not having to install `lerna` globally. But if we go that route, this will have to go with a number of fixes/cleanup to dependencies in the top-level `package.json` along with the `postinstall` change + corresponding fixes to the documentation + changes to CI and maybe deploy scripts. This also will have to be applied systematically to all repos based on lerna.

So I think it would be worth having an issue about this first so others can comment.

PR #112, opened 2019-11-07
chore(package): remove vulnerable unused packages

irmerk:
Signed-off-by: irmerk <jolenelanglinais@gmail.com>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "markdown-transform",
-	"version": "0.7.1",
+	"version": "0.7.3",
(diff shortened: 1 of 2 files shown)


irmerk:
I'm not quite sure, I removed those dependencies and ran `npm install` in `markdown-slate`, then `lerna bootstrap` in the root.

jeromesimeon:
> I'm not quite sure, I removed those dependencies and ran `npm install` in `markdown-slate`, then `lerna bootstrap` in the root.

I think I got confused by the GitHub diff... looks good to me.

Issue #44, opened 2019-10-03
Slate transform does not roundtrip when the text uses softbreaks


jeromesimeon:
**Describe the bug**
Translating from markdown to slate and back sometimes introduces additional new lines.

**To Reproduce**

For instance on the attached sample.md:
[sample.txt](https://github.com/accordproject/markdown-transform/files/3687219/sample.txt)

```sh
$ cat sample.md 
## Fixed rate loan

This is a _fixed interest_ loan to the amount of 100000
at the yearly interest rate of 2.5%
with a loan term of 15,
and monthly payments of {{I'm not sure which amount right now}}
$ markus normalize --sample sample.md --slate
12:34:44 - info: 
Fixed rate loan
----

This is a *fixed interest* loan to the amount of 100000

at the yearly interest rate of 2.5%

with a loan term of 15,

and monthly payments of {{I'm not sure which amount right now}}
```

jeromesimeon:
This issue is most likely related to softbreaks which gets translated into paragraphs. See also https://github.com/accordproject/template-studio-v2/issues/147

dselman:
I think you are correct: https://github.com/accordproject/markdown-transform/blob/master/packages/markdown-slate/src/ToSlateVisitor.js#L235

jeromesimeon:
I'm not sure what is the right way to address this. A couple of possible directions:
1. Preserve softbreaks in the Slate DOM and delay the new line creation until rendering (this might be useful if we want to allow users to control that when editing)
2. Turn softbreaks into a new line folded in the enclosing paragraph (a bit like the HTML transform), but that means we need to recover those

As I'm writing it down, I think like option 1. better but it might be more work (since this is a new kind of Slate node to handle in the editor)... There might be other options to handle this that I'm not thinking about.

dselman:
Yes, I think I like (1) better as well. E.g. is would allow CTRL-ENTER in the editor to insert a softbreak.

FYI, here is what the CommonMark spec says: https://spec.commonmark.org/0.22/#soft-line-breaks

jeromesimeon:
> Yes, I think I like (1) better as well. E.g. is would allow CTRL-ENTER in the editor to insert a softbreak.
> 
> FYI, here is what the CommonMark spec says: https://spec.commonmark.org/0.22/#soft-line-breaks

Yes. That makes sense. @mttrbrts encountered this as well on the HTML side when he looked at pdf rendering, so he may have some thoughts?

[...]

Issue #100, opened 2019-10-28
What should be done with linebreaks/softbreaks in Slate transform?


jeromesimeon:
I'm unclear that this is what should be happening?
https://github.com/accordproject/markdown-transform/blob/90cf0cfa38582e9a38029573ad7fddba8470f32c/packages/markdown-slate/src/ToSlateVisitor.js#L228-L241

dselman:
Interesting thread on this, including when to use <br/>
https://github.com/ckeditor/ckeditor5/issues/4542

jeromesimeon:
Note on this thread:
- Line breaks are two spaces followed by a new line within a paragraph
```
This is a line break  
with something after it
```
- Softbreaks are just a new line
```
This is a softbreak
With something after it
```

In Dingus:
<img width="1378" alt="Screenshot 2019-11-07 at 2 01 14 PM" src="https://user-images.githubusercontent.com/670099/68418847-155e9200-0167-11ea-9556-7bcef2599095.png">

dselman:
Resolved: linebreaks and soft breaks are both represented as inlines in the Slate DOM.

PR #113, opened 2019-11-07
(fix) handle softbreaks and linebreaks in Slate

dselman:

# Issue #100 
Map line break and soft break as inlines in the Slate DOM.

### Changes
Update the Slate transform to handle CommonMark line break and soft break.


jeromesimeon:
This PR also fixed #44

PR #114, opened 2019-11-07
chore(deps): move slate to peerDep and devDep for tests

DianaLease:
Signed-off-by: Diana Lease <dianarlease@gmail.com>
- removes slate from dependencies
- adds slate as a peer dependency
- also adds slate as a dev dependency since it is needed for the tests

### DIFF
--- a/packages/markdown-slate/package.json
+++ b/packages/markdown-slate/package.json
@@ -71,14 +71,17 @@
     "mocha": "6.1.4",
     "nyc": "14.1.1",
     "raw-loader": "^3.0.0",
+    "slate": "^0.47.8",
     "tsd-jsdoc": "^2.3.0",
     "webpack": "^4.35.0",
     "webpack-cli": "^3.3.5"
   },
   "dependencies": {
     "@accordproject/markdown-cicero": "0.7.4",
-    "immutable": "^4.0.0-rc.12",
-    "slate": "^0.47.7"
+    "immutable": "^4.0.0-rc.12"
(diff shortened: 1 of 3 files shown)


packages/markdown-slate/package.json
```diff
@@ -71,14 +71,17 @@
...
+    "slate": "^0.47.8",
     "tsd-jsdoc": "^2.3.0",
     "webpack": "^4.35.0",
     "webpack-cli": "^3.3.5"
   },
   "dependencies": {
     "@accordproject/markdown-cicero": "0.7.4",
-    "immutable": "^4.0.0-rc.12",
-    "slate": "^0.47.7"
+    "immutable": "^4.0.0-rc.12"
```
jeromesimeon: Should `immutable` be also a `peerDependencies` ? What's the reason it is a `peerDependencies` in Slate itself?

DianaLease: I think we can move it to a peerDependencies . Since any app using `markdown-slate` is already going to need to be importing `slate`, which means it will need have a dependency on immutable since it's a peer dep of slate. I'll make this change to match what is in slate.

jeromesimeon: We should probably document this in the package. Slate does mention those in their installation guidelines https://docs.slatejs.org/walkthroughs/installing-slate

DianaLease: Thanks, added it to the readMe

Issue #109, opened 2019-10-30
Issues with Lists in Slate Transform


jeromesimeon:
**Describe the bug**
The slate transform seems to swallow lists in some cases. Here is a sample test.

[list-slate.zip](https://github.com/accordproject/markdown-transform/files/3790312/list-slate.zip)

**update** Markdown to Slate seems to lose any structure except for the first paragraph:
```
bash-3.2$ cat list.md 
1. item one
2. item two
   - sublist
   - sublist

bash-3.2$ markus parse --sample list.md --slate --output data.json 
13:16:40 - info: Creating file: data.json
13:16:40 - info:
{
  "object": "value",
  "document": {
    "object": "document",
    "data": {
    },
    "nodes": [
      {
        "object": "block",
        "type": "ol_list",
        "data": {
          "tight": "true",
          "start": "1",
          "delimiter": "period"
        },
        "nodes": [
          {
            "object": "block",
            "type": "list_item",
            "data": {
            },
            "nodes": [
              {
                "object": "text",
                "text": "item one",
                "marks": []
              }
            ]
          }, 
          {
            "object": "block",
            "type": "list_item",
            "data": {
            },
            "nodes": [
              {
                "object": "text",
                "text": "item two",
                "marks": []
              }
            ]
          }
        ]
      }
    ]
  }
}
```

Which means normalization returns only the top level list:
```
bash-3.2$ markus draft --data data.json --slate 
13:18:14 - info: 
1. item one
2. item two
bash-3.2$ markus normalize --sample list.md --slate
13:18:31 - info: 
1. item one
2. item two
```

This is true of anything which isn't pure text from a single paragraph (breaks, nested lists, multiple chunks of text, etc.)

jeromesimeon:
Proposed fix in #123 

Here is a test with nested lists:
```
bash-3.2$ cat test/data/nested-list.md 
Prolog
1. item one
2. item two
   - sublist
   - sublist

Epilog
bash-3.2$ ../markdown-cli/index.js normalize --sample test/data/nested-list.md --slate
16:45:40 - info: 
Prolog
1. item one
2. item two
   - sublist
   - sublist

Epilog
```

jeromesimeon:
Fixed in #123

PR #123, opened 2019-11-11
feature(slate) Lists in Slate now preserve structured content inside…

jeromesimeon:
… of the list items

# Issue #109 
This revises the behaviour of the Slate transform for lists.

Previously it only took the content of the first paragraph and discarded the rest in the `list_item` Slate node.

With this PR, it constructs any arbitrary Slate DOM within the list based on the Cicero DOM (e.g., paragraphs mapped to paragraphs, nested lists to nested lists, etc).

# Review help

It is unclear if this should be considered a bug fix or a breaking change. This definitely changes the behaviour for single paragraphs (which now have one extra Slate `paragraph` node inside list items).

# Other context

I think this could actually help with the issue here: https://github.com/accordproject/markdown-editor/issues/187
and the corresponding PR here: https://github.com/accordproject/markdown-editor/pull/188

Notably the schema should now assume the content of a list item would have the proper paragraph block -- as suggested by @dselman in comment.


irmerk:
This makes sense to me. I think we would want to have this softbreak example as a test case.

jeromesimeon:
> This makes sense to me. I think we would want to have this softbreak example as a test case.

Great idea. I'll add a test for that.

jeromesimeon:
> This makes sense to me. I think we would want to have this softbreak example as a test case.

Added a test in https://github.com/accordproject/markdown-transform/pull/123/commits/45532a7baac213fde67294040b0cebbd446e148c

irmerk:
As for type, I would think this is a fix rather than a breaking change.

jeromesimeon:
> As for type, I would think this is a fix rather than a breaking change.

I'm worried some previous code relied on the fact that there was no paragraph node in the list item?

PR #133, opened 2019-11-27
feat(markdown-slate): push empty paragraph onto slate dom

irmerk:
# No Issue

### Changes
- Create an empty Slate DOM paragraph block and to tack onto any transform from CiceroMark

### Flags
- Still will require Slate work to handle runtime editing and adding a template to the end of a contract

### Related Issues
- N/A

### DIFF
--- a/packages/markdown-slate/src/SlateTransformer.js
+++ b/packages/markdown-slate/src/SlateTransformer.js
@@ -51,6 +51,20 @@ class SlateTransformer {
             object: 'value',
             document: parameters.result
         };
+        const paragraphSpaceNodeJSON = {
+            object: 'block',
+            type: 'paragraph',
+            data: {
+            },
+            nodes: [
+                {
+                    object: 'text',
+                    text: '',
+                    marks: []
+                }
+            ]
+        };
+        result.document.nodes.push(paragraphSpaceNodeJSON);
         return Value.fromJSON(result);
     }
 
--- a/packages/markdown-slate/src/SlateTransformer.js
+++ b/packages/markdown-slate/src/SlateTransformer.js
@@ -51,20 +51,26 @@ class SlateTransformer {
             object: 'value',
             document: parameters.result
         };
-        const paragraphSpaceNodeJSON = {
-            object: 'block',
-            type: 'paragraph',
-            data: {
-            },
-            nodes: [
-                {
-                    object: 'text',
-                    text: '',
-                    marks: []
-                }
-            ]
-        };
-        result.document.nodes.push(paragraphSpaceNodeJSON);
+        const lastNodeType = result.document.nodes[result.document.nodes.length - 1]
+            ? result.document.nodes[result.document.nodes.length - 1].type
+            : null;
+
+        if (lastNodeType === 'clause') {
+            const paragraphSpaceNodeJSON = {
+                object: 'block',
+                type: 'paragraph',
+                data: {
+                },
+                nodes: [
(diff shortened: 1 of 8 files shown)


packages/markdown-slate/src/SlateTransformer.js
```diff
@@ -51,6 +51,20 @@ class SlateTransformer {
...
+            },
+            nodes: [
+                {
+                    object: 'text',
+                    text: '',
+                    marks: []
+                }
+            ]
+        };
+        result.document.nodes.push(paragraphSpaceNodeJSON);
```
dselman: If you only added this when the last block was a Clause then you wouldn't need to change all the tests... ;-)

packages/markdown-slate/src/slateToCiceroMarkDom.js
```diff
@@ -32,6 +32,19 @@ function slateToCiceroMarkDom(document) {
     // convert the value to a plain object
     const json = JSON.parse(JSON.stringify(document));
     _recursive(result, json.nodes);
+    const isLastNodePara = result.nodes[result.nodes.length - 1]
```
jeromesimeon: This should be made a bit more defensive so it does not fail when the array is empty.

jeromesimeon: Also this should only apply when the last paragraph is preceded by a clause node.

jeromesimeon: Minor code style: `result.nodes[result.nodes.length - 1]` can be put in a variable, somewhat more concise.

packages/markdown-slate/src/slateToCiceroMarkDom.js
```diff
@@ -32,6 +32,19 @@ function slateToCiceroMarkDom(document) {
...
+        .$class === 'org.accordproject.commonmark.Paragraph';
+    const isPragraphEmpty = result.nodes[result.nodes.length - 1].nodes
```
jeromesimeon: Minor code style: `isParagraphEmpty` ?

packages/markdown-slate/src/slateToCiceroMarkDom.js
```diff
@@ -32,6 +32,19 @@ function slateToCiceroMarkDom(document) {
...
+        ? result.nodes[result.nodes.length - 1].nodes[0].text === ''
+        : null;
+    if (isLastNodePara && isPragraphEmpty) {
+        const removeByIndex = (array, index) => array.filter((_, i) => i !== index);
```
jeromesimeon: Maybe would look nicer with: `result.nodes.splice(0, result.nodes.length-1);` ?

packages/markdown-slate/src/slateToCiceroMarkDom.js
```diff
@@ -17,6 +17,33 @@
 const NS = 'org.accordproject.commonmark';
 const NS_CICERO = 'org.accordproject.ciceromark';
 
+const lastNode = (input) => input.nodes[input.nodes.length - 1];
+
+const isSecondLastClause = (input) => input.nodes[input.nodes.length - 2]
```
jeromesimeon: This call will still raise an `undefined` error if there are less than two nodes in the input.
I think those would be best written as `input.nodes[input.nodes.length - 1] && input.nodes[input.nodes.length - 1].$class` etc.

PR #138, opened 2019-12-05
feat(markdown-html): add rules for converting html to ciceroMark

DianaLease:
### Flags
- will unordered lists never have delimiter value? adding it broke the test
- changed from `\n` to `<wbr>` for softbreaks. otherwise I had no way to identify them in the html. Not sure if this will cause issues somewhere else or if there is a better way
- removed some (unnecessary?) `\n`s from the html string visitor to fix new line issues
- still having some new line / white space issues in acceptance.md test
- html inlines and html blocks not yet implemented


DianaLease:
should we remove the `format` parameter for `toCiceroMark` and always return the result as JSON? @jeromesimeon I remember speaking to you about this last week, but I can't remember exactly where that conversation led. Right now that parameter is not doing anything.

jeromesimeon:
> should we remove the `format` parameter for `toCiceroMark` and always return the result as JSON? @jeromesimeon I remember speaking to you about this last week, but I can't remember exactly where that conversation led. Right now that parameter is not doing anything.

Probably a good idea, but would be good to be consistent across all the transformers. I don't think this is a priority right now.

PR #139, opened 2019-12-06
feat(markdown-slate): handle empty paragraphs between clauses

irmerk:
# No Issue
```
Slate => CiceroMark: Strip all empty paragraphs
CiceroMark => Slate: Inject empty paragraph between clauses 
  (and after final clause if it is the end of document)
```

### Changes
- Insert empty paragraph between two clause nodes when transforming to Slate
- Remove all empty paragraphs when transforming to `CiceroMark`

### Flags
- Now all empty paragraphs will be removed when transforming to `CiceroMark`, and thus will not be present in `HTML` or `CommonMark`

### Related Issues
- Pull Request #133


packages/markdown-slate/src/slateToCiceroMarkDom.js
```diff
@@ -19,18 +19,23 @@ const NS_CICERO = 'org.accordproject.ciceromark';
...
+        '$class': 'org.accordproject.commonmark.Paragraph',
+        nodes: [
+            {
+                '$class': 'org.accordproject.commonmark.Text',
+                text: ''
+            }
+        ]
+    };
+
+    input.nodes = input.nodes.filter(node => node !== emptyParagraph);
```
jeromesimeon: You cannot do equality like this in JavaScript. This will _not_ check that the node has the same structure as an empty paragraph, but whether the two nodes are the same object in memory.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators

packages/markdown-slate/src/SlateTransformer.js
```diff
@@ -51,24 +52,39 @@ class SlateTransformer {
...
+                    marks: []
+                }
+            ]
+        };
+
+        // Find any clauses next to each other, force in a paragraph between
+        const newArray = [];
+        result.document.nodes.forEach((node, i) => {
+            newArray.push(node);
+            if (node.type === CLAUSE || result.document.nodes[i + 1] === CLAUSE) {
```
jeromesimeon: - What happens on the last node with `i+1` ?
- Should this compare against the type for the next node as well?

packages/markdown-slate/src/slateToCiceroMarkDom.js
irmerk: Ah yes! That was a weird mistake. I could `JSON.stringify` the two nodes and compare since they will be in the same order.

packages/markdown-slate/src/SlateTransformer.js
irmerk: Another pretty bad mistake, I think I should have made this `&&` rather than `||`.

I should change this to a traditional for loop to handle only going to `array.length - 1`

packages/markdown-slate/src/slateToCiceroMarkDom.js
jeromesimeon: How do we know they will be in the same order?

jeromesimeon: Wouldn't the deleted code do the right thing?

Issue #143, opened 2019-12-12
PDF -> CiceroMark


dselman:
It would be useful to be able to import contracts from existing PDFs.

https://github.com/jzillmann/pdf-to-markdown
https://pandoc.org

This should be integrated with #145

jeromesimeon:
@dselman Can we consider this issue closed with #152?

PR #152, opened 2020-01-02
(feat) convert a PDF to ciceromark

dselman:

# Issue #143 
Code to convert a PDF to CiceroMark DOM

### Flags
- include page numbers for each page of the PDF. I think we need to fork https://github.com/luochen1990/nodejs-easy-pdf-parser to remove
- not tested in the browser


dselman:
> LGTM! One question - this seems like we won't retain any formatting from the PDF document since we are just parsing it into plain text, correct?

Correct.

PR #155, opened 2020-01-27
fix(markdown-html): updates softbreaks

DianaLease:
Signed-off-by: Diana Lease <dianarlease@gmail.com>

### Issue
- Currently, the transform uses `wbr` for softbreaks, which is not semantically correct and causes issues when rendering html in the browser

### Changes
- Uses `\n` for softbreaks instead of `wbr` when converting to html
- When converting from html, we find each `\n` inside of a text node and insert a softbreak node in its place

### Flags
- All of the tests currently pass. However, this could use more testing to make sure extraneous softbreaks do not get added where they shouldn't since we use `\n` in other parts of the transform as well.

Related to https://github.com/accordproject/cicero-template-library/pull/313

### DIFF
--- a/packages/markdown-html/src/ToHtmlStringVisitor.js
+++ b/packages/markdown-html/src/ToHtmlStringVisitor.js
@@ -138,7 +138,7 @@ class ToHtmlStringVisitor {
             parameters.result += '<br>\n';
             break;
         case 'Softbreak':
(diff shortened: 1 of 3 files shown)


jeromesimeon:
Another benefit of the PR is that \n in html paragraphs will now be nicely imported as softbreaks, which could be useful in some cases.

Issue #154, opened 2020-01-08
Issue roundtripping multiline block quotes


DianaLease:
When saving from slate to markdown and then re-rendering as Slate, multiline block quotes get lost. The first line is preserved, but the subsequent lines fall outside the block quote when we reload.

jeromesimeon:
I was mistaken that this isn't a bug in `markdown-transform`. This even occurs for multi-line block quotes roundtrips on CommonMark:

```
bash-3.2$ ./packages/markdown-cli/index.js normalize --sample test.md 
./packages/markdown-cli/index.js normalize --sample test.md 
2:52:27 PM - info: 
test

> code
block

> with

more
than

one
line
test
bash-3.2$ cat test.md 
cat test.md 
test
> code
> block
> > with
> > 
> > more
> than
> 
> one
> line
test
```

jeromesimeon:
There are more issues lurking in the markdown transform around blockquotes, which can in fact contain any markdown.

Here is a sample that shows issues with headings and lists:
```
bash-3.2$ cat ~/test.md
> This is a\
> linebreak.
> This is a quote.
>
> Heading Two
> ----
> This is more text.
> Ordered lists:
> 1. one
> 2. two
> 3. three
bash-3.2$ markus --version
0.9.6
bash-3.2$ markus normalize --sample ~/test.md 
12:08:00 PM - info: 
> This is a\
linebreak.
> This is a quote.
> Heading Two
----
> This is more text.
> Ordered lists:
1. > one
2. > two
3. > three
```

jeromesimeon:
Should be properly fixed in #165 
```
bash-3.2$ cat test.md 
> This is a\
> linebreak.
> This is a quote.
>
> Heading Two
> ----
> This is more text.
> Ordered lists:
> 1. one
> 2. two
> 3. three
bash-3.2$ ../markdown-cli/index.js normalize --sample test.md
11:14:10 PM - info: 
> 
> This is a\
> linebreak.
> This is a quote.
> 
> Heading Two
> ----
> 
> This is more text.
> Ordered lists:
> 1. one
> 2. two
> 3. three
```

jeromesimeon:
PR merged with the fix. The bigger test:

```
bash-3.2$ ../markdown-cli/index.js normalize --sample test/data/neestedblockquote.md 
8:44:21 AM - info: 
Once upon a time
> 
> This is a\
> linebreak.
> This is a quote.
> 
> Heading Two
> ----
> 
[...]

PR #158, opened 2020-02-05
fix(common) Add logic to handle more multilines and nested blockquote…

jeromesimeon:
… content

# Issue #154 

- Tracks block quote level during translation to Markdown
- Adjust logic to handle node transition to handle block quotes

### DIFF
--- a/packages/markdown-common/src/ToMarkdownStringVisitor.js
+++ b/packages/markdown-common/src/ToMarkdownStringVisitor.js
@@ -60,11 +60,11 @@ class ToMarkdownStringVisitor {
     }
 
     /**
-     * Set parameters for inner node
+     * Set parameters for block quote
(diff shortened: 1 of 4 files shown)


packages/markdown-common/src/ToMarkdownStringVisitor.js
```diff
@@ -63,11 +64,12 @@ class ToMarkdownStringVisitor {
      * @param {*} parametersOut - the current parameters
      * @return {*} the new parameters with first set to true
```
DianaLease: should this documentation now say `set to false` instead?

jeromesimeon: oh yes. let me fix the comment. That comment didn't make much sense anyway since that was only used for block quotes and not generally working.

Issue #211, opened 2020-04-10
CLI changes: wrapVariables and noQuote options


dselman:
This is a proposal to make some changes in how we describe wrapVariables and noQuote, and the CLI that drives them.

## wrapVariables

The intention of the transformation option to wrap or unwrap variables is to provide a format that is easier for the Cicero-UI editor to deal with, in that model information for variables (type and formatting information, values for conditional variables) has been moved from the model for the template, inline into the mark-up.

I therefore propose that we define a *new* format, called *CiceroEdit* for this explicit purpose. I think this has several benefits:

- IMO it is clearer what the intention is, and we can provide more complex transformations from `CiceroMark <-> CiceroEdit`, rather than dealing with this using options.

- It is more future proof, as *CiceroEdit* is likely to evolve significantly as we add new capabilities to the editor. For example, when we implement pop-ups for `enums`, we will have to inline the enum values. Having a **name** for this format will help communication.

- The name is intended to signify that this is **not** intended to be a long-lived persistence format, but rather is a transient format used during the scope of a relatively short-lived editing session.

## CLI Commands and noQuote

The `noQuote` options (which removes `"` and `{{` from CiceroMark, rendering it unparseable, is a lossy transformation. I therefore propose that this become an output format in its own right in the CLI, and that we simply refer to it as `--format markdown` rather than as a drafting option. It is then conceptually a peer of the plain text transformation, which should also become an output format.

So markus CLI becomes more like:

```
markus transform --input <file> --from <input-format>  --to <output-format> --output <path to the output file>
```

Formats:
1. markdown
1. commonmark
1. plaintext
1. ciceromark
1. ciceroedit
1. pdf
1. docx 
1. contractdata
1. html
1. slate

Examples

```
markus transform --input sample.md --from commonmark --to ciceromark --output output.md
markus transform --input sample.md --from ciceromark --to plaintext --output output.txt
markus transform --input sample.md --from ciceromark --to ciceroedit --output edit.md
markus transform --input sample.md --from ciceromark --to markdown --output unparseable.md
markus transform --input sample.html --from html --to pdf --output output.pdf
markus transform --input data.json --from json --to pdf --output output.pdf
markus transform --input sample.md --from ciceromark --to contractdata --output data.json
markus transform --input sample.md --from commonmark --to contractdata --output data.json // fails!
markus transform --input sample.md --from markdown --to contractdata --output data.json // fails!
[...]

PR #213, opened 2020-04-13
(feat) New transformation API

dselman:

# Issue #211 
A new API to automatically chain transformation between a source format and a destination format.

### Changes
- Added `dijkstrajs` to find the shortest path between a source and destination format
- Added transformation graph
- Added code to run all the transformations returned by `dijkstrajs`
- Added code to convert the transformation graph to PlantUML

transformations

### Flags
- This code is in the wrong place. It should be somewhere other than `markdown-cli`. Suggestions please!
- Need to add more tests, once we decide where this code should live
- Need detail review of each transformation
- Note that I added `ciceromark_noquotes` as a format, You can "go via" a format by supplying a list of destinations to the `transform` API.

Converts from `ciceredit` to `slate`, going via the 'ciceromark_noquotes' format:

```
        const sample = fs.readFileSync(path.resolve(__dirname, 'data', 'acceptance.md'), 'utf8');
        const result = transform(sample, 'ciceroedit', ['ciceromark_noquotes','slate']);
        result.object.should.equal('value');
```


dselman:
I propose we create a new package under the `markdown-transform` project called `@accordproject/transform`.

dselman:
@jeromesimeon let's talk about how to integrate (if we should integrate...) the transforms which require access to `Template.parse`. It seems like a shame to encumber the `transform` package with all of Cicero, if people only want to convert markdown to html or Slate.

dselman:
Closed in favour of: https://github.com/accordproject/markdown-transform/pull/214

PR #214, opened 2020-04-15
Transform API

dselman:

# Issue #211 
A new API to automatically chain transformation between a source format and a destination format.

### Changes
- Added `dijkstrajs` to find the shortest path between a source and destination format
- Added transformation graph
- Added code to run all the transformations returned by `dijkstrajs`
- Added code to convert the transformation graph to PlantUML and a build script to auto generate the diagram for the readme
- Updated README.md with example usage

transformations


dselman:
> I love it. (Including the fun script to generate the diagram).
> 
> Two things:
> 
> * please remove the `package-lock.json` file

Oops. Will do.

> * I'm slightly uncomfortable having two CLIs. I would lean toward merging but would want the CLI to be consolidated. I am assuming we could move the CLI back into `markdown-cli`. Any thoughts on this?

Agreed - we should refactor the existing CLI to use the new `transform` API. I hope that will clean up the CLI code significantly!

dselman:
@jeromesimeon ready to go

PR #220, opened 2020-04-24
fix(markdown-slate): normalize inlines

irmerk:
Normalize inlines to be surrounded by text nodes

### Changes
- Add a blank text node between inlines, and before or after if an inline is first or last child
- Roundtrip and cleanup empty text nodes from Slate

### Flags
- Removed `Error` throw for `Failed to process node...`

### Related Issues
- [Issue #378](https://github.com/accordproject/cicero-ui/issues/378)


packages/markdown-slate/src/__snapshots__/SlateTransformer.test.js.snap
```diff
@@ -236,104 +236,26 @@ Object {
...
-                  "id": "deliverable",
-                },
-                "object": "inline",
-                "type": "variable",
-              },
               Object {
                 "object": "text",
                 "text": " satisfies the Acceptance Criteria, and ",
               },
-              Object {
```
jeromesimeon, line 290: Something is going seriously wrong in those tests. All the inlines are disappearing. I am suspecting because the logic doesn't push the inlines in the new children list. Input children should always be pushed to the result.

Issue #231, opened 2020-06-19
Escaped markdown doesn't roundtrip


dselman:
`\______________` doesn't roundtrip through the CommonMarkTransformer

```
cat escaped-underscore.md 
\______________Dan-MacBook-Pro-2:data dselman$ markus transform --input escaped-underscore.md --from markdown --to commonmark --verbose --roundtrip
16:57:53 - info: transform input escaped-underscore.md file
16:57:53 - info: transform input escaped-underscore.md printing intermediate transformations.
Converted from markdown to commonmark. Result:
{
  "$class": "org.accordproject.commonmark.Document",
  "xmlns": "http://commonmark.org/xml/1.0",
  "nodes": [
    {
      "$class": "org.accordproject.commonmark.Paragraph",
      "nodes": [
        {
          "$class": "org.accordproject.commonmark.Text",
          "text": "______________"
        }
      ]
    }
  ]
}
Converted from commonmark to markdown. Result:
______________
16:57:53 - info: 
______________
```

jeromesimeon:
Should work now in the new parser with #231

Tests with commonmark roundtrip:
```
bash-3.2$ more test/data/escapes.md
\___
bash-3.2$ ../markdown-cli/index.js transform --input test/data/escapes.md --roundtrip 
1:19:30 PM - info: 
\___
```

Test with parser:
```
bash-3.2$ more test/data/testUnderscore/grammar.tem.md 
{{signatureAnchorTag}}

\__________________

Signature

{{nameAnchorTag}}

\__________________

Name

Date: {{dateAnchorTag}}
bash-3.2$ more test/data/testUnderscore/sample.md 
"signature-clause-signature"

\__________________

Signature

"signature-clause-name"

\__________________

Name

Date: "signature-clause-date"
bash-3.2$ ../markdown-cli/index.js transform --from markdown --to data --grammar ./test/data/testUnderscore/grammar.tem.md --ctoFiles ./test/data/testUnderscore/model.cto --input ./test/data/testUnderscore/sample.md 
1:21:16 PM - info:
{
  "$class": "io.clause.signature.namedate.SignatureClause",
  "signatureAnchorTag": "signature-clause-signature",
  "nameAnchorTag": "signature-clause-name",
  "dateAnchorTag": "signature-clause-date",
  "clauseId": "92f93ec2-f8a2-4ffe-8bef-6b1e3b0740ac"
}
[...]

PR #232, opened 2020-06-19
Escape for underscore, Add escaping to new parser, Refactoring/cleanup

jeromesimeon:
# Issue #231 
- Fixes escaping bug for `_` when serializing text nodes
- Add escaping logic in parser consistent with that used in `markdown-common` for serialization
- Quite a bit of clean up of commonmark and parser generation rules


packages/markdown-common/src/CommonMarkUtils.js
```diff
@@ -161,8 +161,9 @@ function mkATXHeading(level) {
 function escapeText(input) {
     return input.replace(/[*`&]/g, '\\$&') // Replaces special characters
         .replace(/^(#+) /g, '\\$1 ') // Replaces heading markers
-        .replace(/^(\d+)\. /g, '$1\\. ') // Replaces ordered lists markers
-        .replace(/^- /g, '\\- '); // Replaces unordered lists markers
+        .replace(/^(\d+)\. /g, '$1\\. ') // Replaces ordered list markers
+        .replace(/^- /g, '\\- ') // Replaces unordered list markers
+        .replace(/^_/g, '\\_'); // Replaces thematic break markers
```
DianaLease, line 166: I'm a little confused on how we know what is needed here. Do we not need a similar line for block quote markers, for example?

PR #233, opened 2020-06-24
TemplateMark: Monetary Amount support, refactoring, formulas calculation

jeromesimeon:
### Changes
- Refactoring to start aligning markdown text creation with parser generation (to make it more robust for markdown)
- Support for monetary amounts formats
- Extends parser manager with a way to register formula computation during drafting


irmerk:
👀

PR #248, opened 2020-07-07
fix(slate) Add Slate transform for optional

jeromesimeon:

### Changes
- Add missing transform from optional to Slate

### Flags
- Interestingly when drafting in the `none` case we need to invent data for the `some` case since it is required in CiceroMark and Slate. To be discussed maybe.


packages/markdown-slate/src/slateToCiceroMarkDom.js
```diff
@@ -341,6 +350,34 @@ function handleConditional(node, isTrue, whenTrue, whenFalse) {
...
+ * @param {*} whenNone the nodes when absent
+ * @returns {*} the ast node
+ */
+function handleOptional(node, hasSome, whenSome, whenNone) {
+    const data = node.data;
+
+    let result = {
+        $class : `${NS_CICERO}.Optional`,
+        name : data.name,
+        nodes: [],
```
DianaLease, line 367: Does this `nodes` array need to get populated with the text nodes?

jeromesimeon: @DianaLease Good question. This happens naturally as part of the main recursion though. So here:
https://github.com/accordproject/markdown-transform/blob/d88b8900fe1e61e3418bf35c64346adc29f75356/packages/markdown-slate/src/slateToCiceroMarkDom.js#L209

Issue #247, opened 2020-07-07
CiceroMark OOXML transform


dselman:
The existing DOCX support is partial and is poor quality with many real-world DOCX files. It would be preferable to have a first-class bidirectional transformation from CiceroMark to/from OOXML.
**Describe the solution you'd like**
Integrate an OOXML <-> CiceroMark transform into the project.

**Describe alternatives you've considered**
We currently use a 3rd-party library to do DOCX -> Markdown transformation, which has a number of issues.

**Additional context**
- http://officeopenxml.com/WPdocument.php
- https://docs.microsoft.com/en-us/office/dev/add-ins/word/create-better-add-ins-for-word-with-office-open-xml
- https://docs.microsoft.com/en-us/openspecs/office_standards/ms-docx/b839fe1f-e1ca-4fa6-8c26-5954d0abbccd

## Mapping Table

Document -> w:document
Paragraph -> w:p
Text -> w:t
Linebreak -> w:br
Softbreak -> w:cr (?)
?? -> w:tab
?? -> w: noBreakHyphen
?? -> w: softHyphen
List -> w:numbering
ListItem -> w:num
Strong -> w:b
Emph -> w:i

dselman:
@DianaLease @irmerk what is the status of this please? Is there something I can do?

irmerk:
The work for supporting this transform is captured in the [`algoo-ooxml`](https://github.com/accordproject/markdown-transform/tree/algoo-ooxml) branch.

@algomaster99 are you able to update on this?

algomaster99:
The branch [`algoo-ooxml`](https://github.com/accordproject/markdown-transform/tree/algoo-ooxml) currently only comprises of `OOXML -> CiceroMark` transformer and it has only been perfected for `acceptance-of-delivery@0.14.0`.

### Currently parsed entities

It transfers the following OOXML entities into CiceroMark:

1. There are two types of `w:p`. One is a heading the other is an actually paragraph. It is decided by the `w:pStyle` attribute.
    ```xml
      <w:p w:rsidR="009D4C12" w:rsidRDefault="009D4C12">
        <w:pPr>
          <w:pStyle w:val="Heading2"/>
        </w:pPr>
        <w:r>
          <w:rPr>
            <w:sz w:val="40"/>
          </w:rPr>
          <w:t>Acceptance of Delivery.</w:t>
        </w:r>
      </w:p>
      ```
      to 
      ```json
      {
[...]

PR #251, opened 2020-07-09
Conversion of OOXML to CiceroMark and vice versa

algomaster99:
# Issue #247 
Add package for converting `OOXML` <-> `CiceroMark`

### Changes
- [x] initialize package
- [ ] add function for converting OOXML to CiceroMark with proper tests
- [ ] add function for converting CiceroMark to OOXML with proper tests


algomaster99:
Oh, I just thought it was supposed to be a different package. One reason I can think of is that `markdown-docx` uses a **path to a DOCX file** which I _think_ will install dependencies which might only work on Node and not on the browser. `Mammoth` works on the web too though.

jeromesimeon:
> Oh, I just thought it was supposed to be a different package. One reason I can think of is that `markdown-docx` uses a **path to a DOCX file** which I _think_ will install dependencies which might only work on Node and not on the browser. `Mammoth` works on the web too though.

I think it would make sense for all transforms to work on input data in memory and not depend on the file system? Maybe I'm missing something there.

Also, it would help to figure out if there are any differences between ooxml or docx before creating a whole new package.

@DianaLease @irmerk @dselman help here please?

dselman:
> Why is this a different package from `docx` ?

I think making this `OOXML` is more descriptive - as technically that is the (open) data format, vs docx which is the Microsoft proprietary file wrapper.

algomaster99:
@dselman But we will only be able to transform OOXML from Word Documents to CiceroMark. Other OOXML (for eg. from PowerPoint, Excel, etc.) will fail as it makes no sense.

jeromesimeon:
> > Why is this a different package from `docx` ?
> 
> I think making this `OOXML` is more descriptive - as technically that is the (open) data format, vs docx which is the Microsoft proprietary file wrapper.

Sounds like an argument for renaming the current package, not creating a new one?

Also, maybe I'm wrong but isn't a docx containing several xml files? (some of which may or may not be ooxml?).

algomaster99:
> Also, maybe I'm wrong but isn't a docx containing several xml files?

@jeromesimeon OOXML file is just a zipped collection of all the XML files the Word contains as written [here](http://officeopenxml.com/).

irmerk:
After discussing on a call, I think it would likely make the most sense to keep the code in `markdown-docx` instead of making a new package. In the future we can change the name if that makes sense... But based on @algomaster99's findings it seems like `docx` is the most descriptive at the moment.

This will get rid of `mammoth` and replace the [current functionality][code] with the implementation Aman creates. 

[code]: https://github.com/accordproject/markdown-transform/blob/master/packages/markdown-docx/src/DocxTransformer.js#L42-L45

irmerk:
@algomaster99 I think we wanted to keep it `markdown-docx` instead of `markdown-ooxml`

packages/markdown-ooxml/src/OoxmlTransformer.js
```diff
@@ -0,0 +1,161 @@
...
+     * @returns {string} the name/id of the variable
+     */
+    getId(elements) {
+        const variableProperties = elements[0]
+        for (const property of variableProperties.elements) {
+            if (property.name === 'w:tag') {
+                return property.attributes['w:val'];
+            }
+        }
+    }
```
irmerk, line 38: Is there a case where `getId()` will not return anything, and you'd want to have a base case after the for loop?

algomaster99:
@irmerk cool, I will try to integrate it there. Once I will do that, I will close this PR.

packages/markdown-ooxml/src/OoxmlTransformer.js
algomaster99: @irmerk I think every variable will have a corresponding ID so it ought to return something. And I am only calling this function when I parse `w:sdt` which are node types for content controls.

PR #293, opened 2020-08-19
fix(markdown-pdf): crash with empty headings

dselman:

### Changes
- Protect the PDF transform from CiceroMark document that contain headings without text


dselman:
Correct.

Issue #288, opened 2020-08-17
Embedded {{#ulist}} end tag terminates parent {{#ulist}}.


martinhalford:
**Describe the bug**
Where a {{#ulist}} is embedded inside a parent {{#ulist}} the first {{/ulist}} end tag terminates both {{#ulist}} items.

**To Reproduce**
Consider the following grammar.
```
{{#ulist commissions}}
Based on a sale price of $ {{salePrice as "0,0"}} the Commission Amount will be $ {{commissionAmount as "0,0"}} 
Calculated as follows:
{{#ulist lineItems}}
{{description}} being $ {{value as "0,0"}}  
{{/ulist}}
{{/ulist}}
```
This renders as follows:
```
-  Based on a sale price of $ 1,650,000 the Commission Amount will be $ 130,000 (including $ 13,000 GST.)
   Calculated as follows:
   -  7% of the property sale between $0 and $500,000 being $ 35,000
   -  8% of the property sale between $500,000 and $1,000,000 being $ 40,000
   -  9% of the property sale between $1,000,000 and $1,500,000 being $ 45,000
   -  A fixed fee of $10,000 for the property sale price over $1,500,000 being $ 10,000

{{/ulist}}
```
Notice the extraneous `{{/ulist}}` at the end?

Removing the second `{{ulist}}` has no effect on the grammar execution when it should.

**Expected behaviour**

Expect that all `{{#ulist}}` tags should have a matching `{{/ulist}}` tag and, therefore, be able have multiple {{#ulist}} elements embedded in parent {{#ulist}} elements.

**Additional context**
Cicero v0.21.3

jeromesimeon:
That's a serious bug in the new parser (most likely originating from the templatemark plugin). I'm labelling this as a critical bug. I'll look at it ASAP.

Thanks for the report!

jeromesimeon:
Should now be fixed. See PR #300

jeromesimeon:
This very issue also affects `markdown-it-cicero` in the case where some contract contains nested clauses. I'm reopening for that reason.

jeromesimeon:
> This very issue also affects `markdown-it-cicero` in the case where some contract contains nested clauses. I'm reopening for that reason.

Filed as #323

PR #300, opened 2020-08-26
fix(template) Handling for nested lists

jeromesimeon:
…own-it plugin

# Issue #288
Adds proper handling for nested blocks (with stack discipline) to `markdown-it-template` plugin.


jeromesimeon:
Initial work in PR #300 

With it, nested lists get parsed in the properly nested way:

Screenshot 2020-08-26 at 3 40 46 PM

PR #308, opened 2020-08-29
fix(pdf-import): improve the quality of PDF import

dselman:

Reimplement the transformation from PDF to CommonMark, using pdf2json, which itself uses pdf.js. The implementation detects paragraphs, bold and italic and has an option to preserve page breaks.

### Changes
- New implementation of PdfTransformer.toCiceroMark

### Flags
- Occasionally some words get merged, this is due to https://github.com/modesty/pdf2json/issues/94

### Related Issues
- Issue #<NUMBER HERE>
- Pull Request #<NUMBER HERE>


packages/markdown-pdf/src/PdfTransformer.js
```diff
@@ -83,12 +82,88 @@ class PdfTransformer {
...
-        return this.ciceroMarkTransformer.fromMarkdown(plainText, format);
+    async toCiceroMark(input, format = 'concerto', options = { paragraphVerticalOffset: 1, preservePages: true }) {
+        return new Promise( (resolve, reject) => {
+            const pdfParser = new PDFParser(null, false);
+            const errorCallback = (errData) => reject(`PDF parsing failed with error ${errData.parserError}`);
+            const conversionCallback = (pdfData) => {
 
+                const document = {
+                    $class : 'org.accordproject.commonmark.Document',
+                    xmlns : pdfData.formImage.Id.Name,
```
jeromesimeon, line 98: This `xmlns` part of the model is rather confusing. Not sure I know what it means in general.

packages/markdown-pdf/src/PdfTransformer.test.js
```diff
@@ -86,8 +129,11 @@ describe('pdf import', () => {
     getPdfFiles().forEach(([file, pdfContent], i) => {
```
jeromesimeon, line 129: This whole thing needs an `await`, or the subsequent `getMarkdownFiles()` needs to be in a `then`.

dselman: Not sure I understand. Those are separate tests. Jest will await tests that return promises automatically.

jeromesimeon: Doesn't the markdown test need to wait for the files to be written? (I may have misunderstood the structure of the tests).

packages/markdown-pdf/src/PdfTransformer.js
dselman: Agreed - the naming is bad. The idea was that this would be some sort of canonical name/id for the document itself.

packages/markdown-pdf/src/PdfTransformer.test.js
dselman: No, the PDFs the tests write out are just to help debugging. There are separate PDFs that are committed that are the input to the subsequent tests.

PR #317, opened 2020-09-04
fix(pdf) Do not preserve page breaks by defaults on import

jeromesimeon:


DianaLease (approved):
LGTM

PR #325, opened 2020-09-14
fix(markdown-html) : handle linebreaks

dselman:
Issue: #326 

`markdown-html` does not map `Linebreak` in CommonMark to/from `<br/>` in HTML.

### Changes
- Adds a new rule and snapshot test for markdown linebreak

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,28 +21,325 @@
 				"urijs": "1.19.1",
 				"uuid": "3.3.2",
 				"winston": "3.2.1"
+			},
+			"dependencies": {
+				"axios": {
+					"version": "0.19.0",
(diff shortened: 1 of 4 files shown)


DianaLease (approved):
LGTM!

jeromesimeon (changes requested):
Looks good! Couple of things:

- I think this:

https://github.com/accordproject/markdown-transform/blob/dc57c3134d17ebfd1d93a5aff4963bb705bb54fb/packages/markdown-html/src/rules.js#L27

Should also be either removed or fixed?

- Please point to this issue: https://github.com/accordproject/markdown-transform/issues/326

jeromesimeon (approved):
LGTM!

Issue #260, opened 2020-07-13
Transform from markdown to PDF fails when image link exists in markdown.


martinhalford:
**Describe the bug**
Transform command throws an unhandled exception when transforming from markdown to pdf if image link contained within markdown document.

**To Reproduce**
1. Include an image link within markdown document. For example...
```
![logo](https://s3-ap-southeast-2.amazonaws.com/images.benext.io/beNext_orange-darkgreen-sig-870x225.png "")
```
2. Execute transform command. For example...
```
markus transform --input sample.md --to pdf --output output.pdf
```
3. See error
```
(node:66215) UnhandledPromiseRejectionWarning: Invalid image: Error: ENOENT: no such file or directory, open 'https://s3-ap-southeast-2.amazonaws.com/images.benext.io/beNext_orange-darkgreen-sig-870x225.png'
Images dictionary should contain dataURL entries (or local file paths in node.js)
(node:66215) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

**Expected behavior**
The image URL is valid markdown syntax. Ideally would expect PDF to be generated with the image rendered inside the PDF document. Alternatively, would expect graceful handling of image links if not supported.

jeromesimeon:
I'm unclear if we want to support external links for images. @dselman @DianaLease any opinion?

martinhalford:
@dselman  @jeromesimeon  @DianaLease 

If external links for images are not to be supported for `markdown-pdf` then the only way to include images within a contract that uses `markdown-pdf` is via embedded images.

This is fine if images are static (such as logos etc.) which can be embedded within the `grammar.tem.md` but not so easy for dynamic images that form part of the contract data. 

Like many contract signing platforms (e.g. like DocuSign), we are embedding signature images into the contract instance in order to provide reasonable evidence that the contract has been legally signed and therefore is legally enforceable (along with some metadata and selfie images).

In order to do this, images must form part of the `data.json` payload, which is fine if the `data.json` is passed as a file argument to the CLI.

However, when using `Cicero Server`, the `data.json` must be passed to the API as part of the HTTP POST request. 

Unfortunately, even with one (1) small base64 encoded PNG image embedded, the HTTP POST request payload becomes too large.

Current workaround is to use `markdown-html` which supports both linked and embedded images.

[...]

PR #327, opened 2020-09-15
Added support for http url images

martinhalford:
Signed-off-by: Martin Halford <martin@benext.io>

# Issue #260 
Added logic to PDF Transformer to retrieve remote images in TemplateMark as an image reference. 

For example:
`Accord Logo`

Logic added to PDF Transformer to walk the document definition and find all `image` objects, and replace the URL strings with base64 encoded `data:image/png` values.

#### markdown-pdf/src/PDFTransformer.js
- findReplaceImageUrls
  - Walk a JSON object and find any image key starting with 'http' and substitute the URL with the Base64 encoded image data.
  - findReplaceImageUrls is a recursive function.

- getRemoteImageData
  - Asynchronously retrieve image as Base64 encoded data from URL.

#### markdown-pdf/test/data/contract.md
- added `Accord` to test data.

### Note
- Could add flag to disable `findReplaceImageUrls` for very, very large markdown documents which may fail with an out-of-memory error.  All current tests pass and `findReplaceImageUrls`.


packages/markdown-pdf/src/PdfTransformer.js
```diff
@@ -18,56 +18,56 @@ const PDFParser = require('pdf2json');
...
 const PdfPrinter = require('pdfmake');
 const ToPdfMakeVisitor = require('./ToPdfMakeVisitor');
+const axios = require('axios');
 const fonts = {
     Courier: {
         normal: 'Courier',
         bold: 'Courier-Bold',
         italics: 'Courier-Oblique',
-        bolditalics: 'Courier-BoldOblique'
+        bolditalics: 'Courier-BoldOblique',
```
dselman, line 31: These sort of automated (?) re-formatting of the code make it much harder to review PRs, as the substantive changes are now buried.

packages/markdown-pdf/src/PdfTransformer.js
```diff
@@ -173,10 +171,40 @@ class PdfTransformer {
...
     async toPdf(input, options, outputStream) {
-
         const printer = new PdfPrinter(fonts);
 
-        if(!input.getType) {
+        // Asynchronously retrieve image as Base64 encoded data from URL.
+        const getRemoteImageData = async (url) => {
+            try {
+                const buffer = await axios.get(url, { responseType: 'arraybuffer' });
+                return ("data:image/png;base64," + buffer.data.toString('base64'));
```
dselman, line 180: How do you know that the buffer returned in a PNG?

packages/markdown-pdf/src/PdfTransformer.js
```diff
@@ -187,120 +215,122 @@ class PdfTransformer {
         const visitor = new ToPdfMakeVisitor(this.options);
         input.accept(visitor, parameters);
 
-        const dd = parameters.result;
-        // console.log(JSON.stringify(dd, null, 2));
+        let dd = parameters.result;
+        await findReplaceImageUrls(dd);
```
dselman, line 219: Let's wrap this in an option (can default to true, so that it can be easily turned off.)

dselman (changes requested):

dselman:
Very nice. A couple of requested changes but the approach looks solid!

martinhalford:
Credit to @dselman - who did most of the heavy lifting by showing me where and how to swap out the image URLs in the document definition.  Thanks Dan!

packages/markdown-pdf/src/PdfTransformer.js
dselman: I'm more worried about what happens with no network access of behind a firewall, than out of memory.

packages/markdown-pdf/src/PdfTransformer.js
martinhalford: Yes - looks like my code prettier made some format changes.  If this is an issue then I can try to revert this.

packages/markdown-pdf/src/PdfTransformer.js
martinhalford: Ummm... I don't. It was just that the raw `buffer.data` was large and thought maybe using base64 png would be nicer.  Could maybe make this an option too?

packages/markdown-pdf/src/PdfTransformer.js
martinhalford: Ok.. Will do.

packages/markdown-pdf/src/PdfTransformer.js
martinhalford: I'll remember to turn off my code prettifier in future. Sorry.

packages/markdown-pdf/src/PdfTransformer.js
martinhalford: I'm thinking it might be better to allow no network access and/or unreachable URLs errors to fail gracefully.  I'm adding logic to insert a placeholder image if the `axios.get` request fails, for whatever reason - like web browser do when there's a broken image link.  We need to do this anyway.  The markdown-pdf transform will fail if we leave unresolved `` inline.

We can add an option to turn off image URL replacement if you like but this would only really be useful if there are no image URLs in the markdown and the user is looking to save processing time when transforming a particularly large file.

packages/markdown-pdf/src/PdfTransformer.js
martinhalford: Ok... Am changing logic to use `Buffer.from(buffer.data)` with no base64 encoding.

Issue #249, opened 2020-07-07
Make decorators on variables available within CiceroMark


dselman:
It would be very useful for transformations to have access to all the decorators placed on variable Concerto types -- allowing them to customise the output based on metadata specified on the variable types.

For example, this could be used by signature blocks to ensure that variables with the `@SignatureVariable()` decoration are rendered as white text in PDF, and are not editable in the Slate editor.

See: https://github.com/accordproject/cicero/issues/557

```
/**
 * A Signature block
 */
asset SignatureClause extends AccordClause {

  @SignatureVariable()
  --> AccordParty party

  @DocuSignTab("Text")
  @SignatureVariable()
  o String favoriteColor

  @DocuSignTab("Title", "optional")
  @SignatureVariable()
  o String title

  @DocuSignTab("SignHere")
  @SignatureVariable()
  o String signature

  @DocuSignTab("FullName")
  @SignatureVariable()
  o String name

  @DocuSignTab("DateSigned")
  @SignatureVariable()
  o String date
}
```

**Describe the solution you'd like**
All decorators on a Concerto type to be copied into the typed CiceroMark variable node.

**Describe alternatives you've considered**
None.

**Additional context**

To access decorators:

```
const val = myField.getDecorators();
```

jeromesimeon:
ouch... I completely forgot about this. I think this will have to happen in a future release.

dselman:
Implemented.

PR #329, opened 2020-09-16
feat(markdown-pdf) : allowing overriding the pdf style using a Concerto decorator

dselman:

Allow the style for PDF elements (which defaults to the concerto type name) to be overridden via a Concerto decorator on the model element associated with a CiceroMark DOM node.

# Issue #330

### Changes
- In the code that generates the pdfmake DOM check whether there is a decorator on the CiceroMark DOM node

### Flags
- <DESCRIBE ISSUES OR HOLDS FOR REVIEWERS>

### Related Issues
- Issue #249


packages/markdown-pdf/src/PdfTransformer.test.js
```diff
@@ -151,6 +163,42 @@ describe('pdf import 2', () => {
     });
 });
 
+describe('pdf generation with decorators', () => {
+    it.only('converts signature-block.md', async () => {
+        const templateTransformer = new TemplateMarkTransformer();
+        const templateMarkContent = fs.readFileSync( path.join(__dirname, '/../test/data/signature', 'grammar.tem.md'), 'utf-8' );
+
+        const modelManager = templatemarkutil.templateMarkManager.modelManager;
```
jeromesimeon, line 171: you should be able to use `const modelManager = await ModelLoader.loadModelManager(null, models, options);` to replace a lot to this, and remove the `AP_SYSTEM_MODEL`.

jeromesimeon (changes requested):

jeromesimeon (approved):
LGTM!

Issue #336, opened 2020-09-21
Tests do not run the linter by default


jeromesimeon:
The top-level tests do not run the linter by default, which means linting issues do not trip the CI on PRs.

For instance, this PR #334 passes the tests, but running the tests in the `markdown-pdf` package fails:
```
bash-3.2$ npm run test 

> @accordproject/markdown-pdf@0.12.10 pretest /Users/jeromesimeon/git/markdown-transform/packages/markdown-pdf
> npm run lint && npm run build


> @accordproject/markdown-pdf@0.12.10 lint /Users/jeromesimeon/git/markdown-transform/packages/markdown-pdf
> eslint .


/Users/jeromesimeon/git/markdown-transform/packages/markdown-pdf/src/PdfTransformer.js
  178:34  error  Strings must use singlequote                                                                            quotes
  195:78  error  Trailing spaces not allowed                                                                             no-trailing-spaces
  196:58  error  Trailing spaces not allowed                                                                             no-trailing-spaces
  201:25  error  Possible race condition: `object[key]` might be reassigned based on an outdated value of `object[key]`  require-atomic-updates
  205:25  error  Possible race condition: `object[key]` might be reassigned based on an outdated value of `object[key]`  require-atomic-updates
  208:29  error  Possible race condition: `object[key]` might be reassigned based on an outdated value of `object[key]`  require-atomic-updates
```

jeromesimeon:
Fixed in #337

Issue #335, opened 2020-09-21
pdf generation doesn't handle complicated headings properly


jeromesimeon:
**Describe the bug**
Transform to pdf does not handle heading with complex content properly, losing some of the content and internal markup.

**To Reproduce**
Try to transform the following markdown to pdf:
```
This is a
*complicated*
heading
==============

Here is some text
```

Using Markus:
```
bash-3.2$ markus transform --from markdown --to pdf --output test.pdf --input test.md 
10:37:32 AM - info: Creating file: test.pdf
10:37:32 AM - info: 
<binary pdf data>
```

This results in the following pdf:
[test.pdf](https://github.com/accordproject/markdown-transform/files/5256050/test.pdf)

The result should be consistent with the rendered markdown:
<img width="1475" alt="Screenshot 2020-09-21 at 10 35 33 AM" src="https://user-images.githubusercontent.com/670099/93780791-93e4e680-fbf6-11ea-900e-b39f0525be95.png">

jeromesimeon:
Fixed in #337 

Fix handles table of content. For instance on the new `test/data/contract.md`

Containing a heading 2:
```
Title2
*(Heading Two)*
**on several lines**
----

```

Properly generates a TOC:

<img width="882" alt="Screenshot 2020-09-21 at 3 16 00 PM" src="https://user-images.githubusercontent.com/670099/93811338-fdc4b680-fc1d-11ea-8fbd-c5f03f27c407.png">

PR #337, opened 2020-09-21
chore(build) Always call linter on testing

jeromesimeon:

# Issue #336 #335 
- Turns on linter for global `npm run test`, ensuring it is run by the CI
- Fixes linting issues for `markdown-pdf` package, including race condition on async/await
- Fixes pdf transform for headings, now can handle various kinds of content/inlines


jeromesimeon:
@martinhalford Would love a review on the second commit if you can.

dselman (approved):
LGTM

martinhalford (commented):
@jeromesimeon. This is a nice improvement. By storing changes in the updates array and then updating the key in the `foreach` loop later, we avoid any issues with erroneous updates.  However, not sure why we are using a `.then()` since we are awaiting the `Promise.all`.

jeromesimeon:
> @jeromesimeon. This is a nice improvement. By storing changes in the updates array and then updating the key in the `foreach` loop later, we avoid any issues with erroneous updates. However, not sure why we are using a `.then()` since we are awaiting the `Promise.all`.

Thanks! I think this is still correct, since the await will apply to the combination of `Promise.all(...).then(...)` but maybe you are right that this could have been written more nicely without that `then()` by placing the result of the `Promise.all` into a variable.

The purpose of the change was essentially to avoid having multiple `object[key] = ...;` in parallel which could result in race conditions.

martinhalford:
> Thanks! I think this is still correct, since the await will apply to the combination of `Promise.all(...).then(...)` but maybe you are right that this could have been written more nicely without that `then()` by placing the result of the `Promise.all` into a variable.
> 
> The purpose of the change was essentially to avoid having multiple `object[key] = ...;` in parallel which could result in race conditions.

You are probably right. Anyway, really good idea to avoid the race conditions. Thanks for fixing that!

PR #351, opened 2021-01-27
feat(pdf) : load/save CiceroMark and templates to PDF

dselman:

### Changes
- Embeds CiceroMark into the PDF metadata
- Embeds an array of templates into the PDF metadata (as base64 strings)
- Loads CiceroMark from the PDF metadata and returns it
- Loads templates from the PDF metadata and returns them (as base64 strings)

pdf-metadata
_Viewing PDF metadata inside Acrobat_

[roundtrip.md.pdf](https://github.com/accordproject/markdown-transform/files/5880632/roundtrip.md.pdf)
_PDF with embedded CiceroMark and Acceptance of Delivery template_

### Flags
- Should we add an optional attribute `templates` to org.accordproject.commonmark.Document?
- What are the size limitations on metadata?
- Should we change the implementation so that the user supplies a map object containing the templates (with template uri as the key, and the template archive as the value), and then we walk the CiceroMark including all the templates that are referenced?
   - the advantage of this approach would be that the recipient of the PDF could "rewire" those CiceroMark clauses after they've uploaded the templates to their private repository

### Related Issues
- https://github.com/accordproject/cicero/issues/562
- Pull Request #<NUMBER HERE>


dselman:
@jeromesimeon for review and discussion. This gets us a step closer to being able to exchange CiceroMark + templates and uses the PDF as the carrier in this case.

jeromesimeon (approved):
LGTM

dselman:
TemplateMark + Data => CiceroMark : Use case 1, allows someone to see the values of the data, and regenerate the natural language

Execution would require Cicero Template : Use case 2, allows someone to simulate or process transactions

```
fn(templateMark) -> template*
templateUri : base64
```

jeromesimeon:
@dselman Can we get a status report on this tomorrow? I'm unclear what proposal went in and if we need to open an issue about this (was there ever one?).

jeromesimeon:
> TemplateMark + Data => CiceroMark : Use case 1, allows someone to see the values of the data, and regenerate the natural language
> 
> Execution would require Cicero Template : Use case 2, allows someone to simulate or process transactions
> 
> ```
> fn(templateMark) -> template*
> templateUri : base64
> ```

Most specifically was this proposal implemented?

dselman:
Sure. The merged code allows the caller to optionally pass an array of templates that are stored in the PDF (off by default).

PR #352, opened 2021-01-29
chore(github): migrate travis ci/cd to gh actions - I4

irmerk:
# https://github.com/accordproject/technical-steering-committee/issues/4
Migrate TravisCI workflows to GitHub Actions

### Changes
- Removes `travis` related code

### Flags
- N/A

### Related Issues
- https://github.com/accordproject/concerto/pull/219


.github/workflows/build.yml
```diff
@@ -0,0 +1,94 @@
...
+      job-status: ${{ job.status }}
+
+    steps:
+      - name: git checkout
+        uses: actions/checkout@v2
+
+      - name: lerna install
+        run: |
+          npm install lerna@^3.22.0
+          ./node_modules/.bin/lerna bootstrap 2>&1
```
jeromesimeon, line 31: Actually we will need to change this one to `npm install`.

irmerk: Changed in `publish.yml` as well.

jeromesimeon: 👍 

The reason is because we do something a little tricky in the markdown transform: before calling `lerna bootstrap` we need to download the latest versions of CiceroMark/TemplateMark coo models (you can see in the `package.json` what happens on install to get an idea.

jeromesimeon (approved):

PR #356, opened 2021-01-29
fix(github): ignore dingus on npm publish

irmerk:
Continues #355


jeromesimeon (changes requested):
No I believe this is a lerna option not an npm option:

lerna exec —ignore dingus — npm ...

PR #367, opened 2021-03-23
fix(markdown-slate): linked variables

d-e-v-esh:
Signed-off-by: d-e-v-esh <59534570+d-e-v-esh@users.noreply.github.com>

# Helps to resolve https://github.com/accordproject/web-components/issues/175

The list items of the `volumediscountolist` template have linked variables. This is because they all are dealing with the same kind of variables. Checking of those variables are inside of a list and if that is true then changing their names fixes this issue.

### Changes

 - Each line in the list comes as an object and encased under the type `list_item`. 
 - They all have a key called `children` that contains an array with one element inside it.
 
chrome_FhfxaXRWCx

That element is another object with the type `paragraph` that contains the data for the actual data that is in the list item.

ZiCAagdCY1

It contains another array called `children` with the data that is inside the line like this.

chrome_kbQnAiimZi

As you can see, it has 7 different objects as elements of the array. Each element is there to create a divide between the normal text and variables and all the variables have a type variable

Each variable contains an object called `data` where the name of the variable resides

chrome_b8MvfbKAaf

We have to change that name and it works. We basically check if a variable is inside a list item and if that is true then we add the index of the list next to the variable and it looks like this.

chrome_auLJtPLeA9

### Author Checklist
- [x] Merging to `master` from `fork:branchname`

### DIFF
--- a/packages/markdown-slate/lib/ToSlateVisitor.js
+++ b/packages/markdown-slate/lib/ToSlateVisitor.js
@@ -499,11 +499,24 @@ class ToSlateVisitor {
             if (thing.decorators) {
                 data.decorators = thing.decorators.map(x => parameters.serializer.toJSON(x));
             }
+
+            const calculatingChildren = this.processChildNodes(thing, parameters);
+            // This changes the variable names if they are inside a list so that they are not linked.
+            for(let i = 0; i < calculatingChildren.length; i++){
+                const listElements = calculatingChildren[i].children[0].children;
+                for(let j = 0; j < listElements.length; j++){
+                    if(listElements[j].type === 'variable'){
+                        const listVariables = `${listElements[j].data.name}${i}`;
+                        listElements[j].data.name = listVariables;
+                    }
+                }
+            }
+
             result = {
                 object: 'block',
                 data: data,
                 type: thing.type === 'ordered' ? 'ol_list' : 'ul_list',
-                children: this.processChildNodes(thing,parameters)
+                children: calculatingChildren
             };
         }
             break;


irmerk:
Just a quick note @d-e-v-esh, this won't exactly close https://github.com/accordproject/web-components/issues/175, because at the very least we'll want to release and update this package within `web-components` (assuming this PR does fix the bug).

d-e-v-esh:
@irmerk That is totally good. I can create another PR for `web-components` whenever needed.

irmerk:
Pinging @jeromesimeon and @dselman on this PR.

jeromesimeon:
hm...that doesn't seem right to me. If you change the variable names in Slate, you won't be able to convert Slate back to its original CiceroMark.

jeromesimeon:
Before you open a PR, it's useful to check that the linter and tests pass locally. You can do that by running `npm run test` in the root directory, or in the specific package you are changing.

d-e-v-esh:
> hm...that doesn't seem right to me. If you change the variable names in Slate, you won't be able to convert Slate back to its original CiceroMark.
> 
> Also this issue isn't specific to lists. It will happen for any nested variable with the same name.

I don't know what would be the best way to solve this issue. This solution was proposed by Dan [here](https://github.com/accordproject/web-components/issues/175#issuecomment-796711799). I followed through and it fixed the issue so I made the PR.

d-e-v-esh:
> Before you open a PR, it's useful to check that the linter and tests pass locally. You can do that by running `npm run test` in the root directory, or in the specific package, you are changing.

Thanks for telling. I didn't know about that. I'll fix this PR and keep that in mind from now.

PR #410, opened 2021-06-16
feat(markdown-docx): add variable transformer - #397

### LINKED ISSUE #397: CiceroMark<->OOXML transformers
# Feature Request 🛍️

Incorporating the CiceroMark->OOXML transformer and improving the currently implemented OOXML->CiceroMark transformer.

## Use Case

It will allow the conversion of Docx files into CiceroMark JSON and vice versa. In addition, using this `Docx` files can be converted into different other formats like `PDF` or `HTML` as well.

## Possible Solution

The transformer for OOXML->CiceroMark is already implemented. It needs to be updated with some entities to allow full transformation.

The transformer for CiceroMark->OOXML is created in `cicero-word-add-in` branch. It needs to be transferred/transported here with some changes to allow the transformation.

## Detailed Description

Currently, the CiceroMark->OOXML transformer supports the following conversions:
| CiceroMark Entity  | OOXML Entity |
| --------  | -------- |
| Text      | `<w:t>` |
| Paragraph     | `<w:p>Content</w:p>` |
| Linebreak      | `<w:p/>` |
| Softbreak      | `<w:r><w:sym></w:r>` |1
| Emph     | `<w:i>` |
| Variable      | `<w:sdt>` |
| List Block/ List     | `<w:numPr><w:num w:val={ordered/unordered}/></w:numPr>` |
| List Item(Text)     | `<w:t/>` |
| List Item(Variable)     | `<w:sdt/>` |

Conversions which are left:

- `Formula / Ergo expressions`
- `Clause`
- `Conditional`
- `Optional`
- `Strong`
- `BlockQuote`
- `CodeBlock`
- `Code`
- `ThematicBreak`
- `Link`
- `Image`

In the left conversions, we need to decide which ones need major importance/priority and which can be given a lower priority. Furthermore, we also need to think about whether all these will be present in the contract (IMO, Code and CodeBlock generally won't occur in the contract)

K-Kumar-01:

Partial fix #397 

Adds the roundtrip transformation for `variable`.

### Changes

- Add helpers and rule
- Add the transformation logic
- Add the test for above

### Author Checklist


packages/markdown-docx/src/CiceroMarkToOOXML/index.js
```diff
@@ -60,6 +60,28 @@ class CiceroMarkToOOXMLTransfomer {
...
+            const tag = node.name;
+            const type = node.elementType;
+            if (Object.prototype.hasOwnProperty.call(counter, tag)) {
+                counter = {
+                    ...counter,
+                    [tag]: {
+                        ...counter[tag],
+                        count: ++counter[tag].count,
+                    },
+                };
```
algomaster99, line 73: No change here. But I just wanted to confirm in you know why `counter` variable is introduced in the first place.

packages/markdown-docx/src/CiceroMarkToOOXML/rules.js
```diff
@@ -78,4 +78,39 @@ const HEADING_RULE = (value, level) => {
...
+ * @returns {string} OOXML string for the variable
+ */
+const VARIABLE_RULE = (title, tag, value, type) => {
+    return `
+      <w:sdt>
+        <w:sdtPr>
+          <w:rPr>
+            <w:color w:val="000000"/>
+            <w:sz w:val="24"/>
+            <w:highlight w:val="green"/>
```
algomaster99, line 97: Remove this for now. We will discuss later if we want to highlight variables or now.

packages/markdown-docx/src/CiceroMarkToOOXML/rules.js
```diff
@@ -78,4 +78,39 @@ const HEADING_RULE = (value, level) => {
...
+          </w:rPr>
+          <w:alias w:val="${titleGenerator(title, type)}"/>
+          <w:tag w:val="${tag}"/>
+        </w:sdtPr>
+        <w:sdtContent>
+          <w:r>
+            <w:rPr>
+              <w:color w:val="000000"/>
```
algomaster99, line 105: I am not sure if this would be necessary as the default colour is black only. Could you confirm?

algomaster99 (changes requested):
Minor changes, then we would be good to go.

I also want you to keep a mental note that the `CiceroMarkToOOXML` transformer should also return the `counter` object as this would be used in MS Word Add-in. If we don't return it, we would have to parse the OOXML string in the add-in to generate it, which is _very_ prone to error. Since this PR works well on its own, you can do that in a separate PR but first, just understand why the `counter` variable is important.

packages/markdown-docx/src/CiceroMarkToOOXML/rules.js
K-Kumar-01: @algomaster99 Yes the default color is black. This can be removed.

packages/markdown-docx/src/CiceroMarkToOOXML/index.js
K-Kumar-01: @algomaster99 
A variable of the same type, say `Shipper` can have multiple instances in a template. The `counter` variable counts the number of instances of that variable. Now using this counter, a title is generated for the variable content control to distinguish it, This is my understanding of the above.

packages/markdown-docx/src/CiceroMarkToOOXML/rules.js
K-Kumar-01: @algomaster99 I will create a PR in the morning.

algomaster99: Sure. Don't create a new one. You can update this one only.

packages/markdown-docx/src/CiceroMarkToOOXML/index.js
algomaster99: > Now using this counter, a title is generated for the variable content control to distinguish it

Almost got it. A more precise answer - each content control in the document must have a unique title.

K-Kumar-01:
@algomaster99 
I have updated the PR. Let me know if there are any changes required.

algomaster99 (approved):
🎉

PR #414, opened 2021-06-18
refactor(markdown-docx): export getClass as helper - #397

K-Kumar-01:

Need: While working on `LIST_RULE`, there was requirement of `node.$class`. Since a function has already been created in transformer, the decision was to export it as a helper and use it.

### Changes

-  use getClass as helper instead of transformer class function

### Author Checklist

### DIFF
--- a/packages/markdown-docx/src/CiceroMarkToOOXML/helpers.js
+++ b/packages/markdown-docx/src/CiceroMarkToOOXML/helpers.js
@@ -96,4 +96,14 @@ function wrapAroundDefaultDocxTags(ooxml) {
     return ooxml;
 }
 
-module.exports = { sanitizeHtmlChars, titleGenerator, wrapAroundDefaultDocxTags };
+/**
+ * Gets the class of a given CiceroMark node.
+ *
+ * @param {object} node CiceroMark node entity
+ * @returns {string} Class of given node
+ */
+function getClass(node) {
+    return node.$class;
+}
+
+module.exports = { sanitizeHtmlChars, titleGenerator, wrapAroundDefaultDocxTags, getClass };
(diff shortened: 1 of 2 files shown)


packages/markdown-docx/src/CiceroMarkToOOXML/helpers.js
```diff
@@ -96,4 +96,14 @@ function wrapAroundDefaultDocxTags(ooxml) {
...
-module.exports = { sanitizeHtmlChars, titleGenerator, wrapAroundDefaultDocxTags };
+/**
+ * Gets the class of a given CiceroMark node.
+ *
+ * @param {object} node CiceroMark node entity
+ * @returns {string} Class of given node
+ */
+function getClass(node) {
+    return node.$class;
+}
```
algomaster99, line 107: I would advise against this because `getClass` is supposed to be _bound_ to the class as you will have access to "node" only in the class, so it doesn't make sense to expose it. Moreover, helper methods don't have such strict parameter requirement. For example, `sanitizeHtmlChars` is a perfect example of a helper method because it can be a static method too as it just needs a string and doesn't matter from where we invocate it. However, `getClass` depends strongly on the input because of the `$class` of `node` and you will only get this while iterating over CiceroMark. So it's better to keep this as an instance method only.

To summarise, helper methods are those functions that are not much dependent upon the state of the application and can be reused throughout the code. You can also think of them as static methods in a class.

algomaster99 (commented):
Although a nice initiative, I disagree with this design. See comment.

packages/markdown-docx/src/CiceroMarkToOOXML/helpers.js
K-Kumar-01: @algomaster99 Shall I close this PR then?

algomaster99: If you are with my stand, then yes.

K-Kumar-01: @algomaster99 Yeah, I think it makes sense but I have one doubt. So we want `node.$class` in the list transformer.
Should any function returns a node.$class or writes it as `node.$class` in the comparison itself?
We currently use `this.getClass` in the transformer itself and the rule will be imported from another file.

algomaster99: Try to decouple rules from the node object. I feel there should be a way.

PR #417, opened 2021-06-23
feat(markdown-docx): add strong transformer - #397

### LINKED ISSUE #397: CiceroMark<->OOXML transformers

K-Kumar-01:

Reference #397

### Changes

- Rules: STRONG_RULE
- Class in definedNodes
- Transformer(CiceroMark JSON <-> OOXML)
- Corresponding Test

### Author Checklist


packages/markdown-docx/src/CiceroMarkToOOXML/rules.js
```diff
@@ -47,6 +47,24 @@ const EMPHASIS_RULE = (value) => {
...
+ *
+ * @param {string}  value  Enclosing value of the OOXML tag
+ * @returns {string} OOXML tag for the strong text
+ */
+const STRONG_RULE = value => {
+    return `
+      <w:r>
+        <w:rPr>
+          <w:b />
+          <w:bCs />
```
algomaster99, line 61: What is this tag for? `<w:b>` is already there. Do we need <w:bCs>` too?

K-Kumar-01: `bCs` is used for complex [script](https://c-rex.net/projects/samples/ooxml/e1/Part4/OOXML_P4_DOCX_bCs_topic_ID0EGNFO.html). 
This is the reason why I have used it.

algomaster99: Are we using complex scripts? All templates are in English only, right?

algomaster99:
Screenshot from 2021-06-23 19-22-29
Don't use the "fix" prefix here. GitHub will close the issue when this PR will be merged.

packages/markdown-docx/src/CiceroMarkToOOXML/rules.js
K-Kumar-01: As far as i remember,  one of the templates used some different language. The template was probably `car rental`. Besides, it might be possible that later templates might be created in different languages. So i guess there would be no harm keeping this.

algomaster99: > Besides, it might be possible that later templates might be created in different languages.

Good point. In that case, we can keep it.

PR #418, opened 2021-06-25
fix: rewrite transformer(CiceroMarkToOOXML) logic to handle nesting - #397

K-Kumar-01:
Signed-off-by: k-kumar-01 <kushalkumargupta4@gmail.com>

### Changes

- Remove old transformer
- Rules: EMPHASIS, TEXT, TEXT_STYLES, TEXT_WRAPPER, PARAGRAPH_WRAPPER, PARAGRAPH_PROPERTIES, STRONG, HEADING
- Tests: Check only for text-and-emphasis using if

### Author Checklist
- [x] Ensure you provide a [DCO sign-off](https://github.com/probot/dco#how-it-works) for your commits using the `--signoff` option of git commit.
- [x] Vital features and changes captured in unit and/or integration tests
- [x] Commits messages follow [AP format](https://github.com/accordproject/techdocs/blob/master/DEVELOPERS.md#commit-message-format)
- [ ] Extend the documentation, if necessary
- [ ] Merging to `master` from `fork:branchname`

PR #421, opened 2021-06-30
fix: rewrite transformer(OOXML->CiceroMark)

K-Kumar-01:
Signed-off-by: K-Kumar-01 <kushalkumargupta4@gmail.com>

This PR deals with rewriting the current OOXML transformer so that it supports nesting.

### Changes 
Major changes will be as follows:
- Use DFS to traverse nodes
- Generate node using the properties stored.
- compare adjacent nodes for common properties(first n) and nest the second one into the common depth.

### Author Checklist


K-Kumar-01:
@algomaster99 @dselman 
I have made the commit for `OOXML Transformer`. The tests will fail as of now. 
The PR is made so that I can get a review on the logic used.

algomaster99:
If the tests fail, there must be something wrong with the logic. Let the tests speak for your code. That's why I stress on rigorous testing. The more unique test cases you add, the better your code becomes.

K-Kumar-01:
> If the tests fail, there must be something wrong with the logic. Let the tests speak for your code. That's why I stress on rigorous testing. The more unique test cases you add, the better your code becomes.

By test fail, I mean that the current transformation is not complete, so for that the test fail. For the conversion `text-and-emphasis`, it passes.

K-Kumar-01:
@algomaster99 @dselman 
Requesting your reviews. The rewriting of the transformer is completed. It passes the previous tests and also the `nesting.json` test which was added in the [PR](https://github.com/accordproject/markdown-transform/pull/418)

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
+     * @return {object} CiceroMark Node
      */
-    getValue(elements) {
-        const variableContents = elements[1];
-        for (const property of variableContents.elements[0].elements) {
-            if (property.name === 'w:t') {
-                return property.elements[0].text;
-            }
+    construtCiceroMarkNodeJSON(nodeInformation) {
+        let obj = {};
```
algomaster99, line 96: I can infer from the method name that you are constructing a CiceroMark node. You should rename this variable accordingly.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
-     */
-    getHeading(headingElement) {
-        if (headingElement.attributes !== undefined) {
-            const headingLevel = headingElement.attributes['w:val'];
-            if (headingElement.name === 'w:pStyle' && headingLevel.includes('Heading')) {
-                return {
-                    isHeading: true,
-                    level: headingLevel[headingLevel.length - 1],
+            let rootNodesLength = 1;
+            for (let i = 1; i < this.JSONXML.length; i++) {
```
algomaster99, line 135: What exactly `i` is? Could you rename it so that it makes more sense?

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
+                name: nodeInformation.name || 'No name available',
+            };
+        } else {
+            obj = {
+                $class: `${NS_PREFIX_CommonMarkModel}Text`,
+                text: nodeInformation.value,
+            };
         }
-        return null;
+        for (let i = nodeInformation.properties.length - 1; i >= 0; i--) {
```
algomaster99, line 114: I think a better name would be `nodePropertyIndex` instead of `i`.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
+     * @param {object} rootNode Pargraph, heading or any other node
      */
-    isLineBreak(element) {
-        return element.name === 'w:p' && element.elements === undefined;
-    }
+    constructNodes(rootNode) {
+        if (this.JSONXML.length > 0) {
+            let constructedNode;
+            constructedNode = this.construtCiceroMarkNodeJSON(this.JSONXML[0]);
+            rootNode.nodes = [...rootNode.nodes, constructedNode];
```
algomaster99, line 132: I am assuming you're getting `rootNode` directly from the OOXML (converted to JSON by `xml-js`). Don't modify it because we don't want any sideffects.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
+                        commonLength++;
+                    } else {
+                        break;
+                    }
+                }
+                let updatedProperties = {
+                    ...this.JSONXML[i],
+                    properties: [...this.JSONXML[i].properties.slice(commonLength)],
                 };
+                constructedNode = this.construtCiceroMarkNodeJSON(updatedProperties);
```
algomaster99, line 151: Could you explain why are you calling this method twice in the same block? Just wanted to understand the functioning.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
-            if (headingElement.name === 'w:pStyle' && headingLevel.includes('Heading')) {
-                return {
-                    isHeading: true,
-                    level: headingLevel[headingLevel.length - 1],
+            let rootNodesLength = 1;
+            for (let i = 1; i < this.JSONXML.length; i++) {
+                let propertiesPrevious = this.JSONXML[i - 1].properties;
+                let propertiesCurrent = this.JSONXML[i].properties;
+
+                let commonLength = 0;
```
algomaster99, line 139: `commonPropertiesLength` would be more appropriate I think.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
-        case 'w:sym':
-            return {
-                $class: `${NS_PREFIX_CommonMarkModel}Softbreak`,
-            };
-        case 'w:r':
-            if (element.elements[0].name === 'w:rPr') {
-                let emphasisedTextFound = element.elements[0].elements.some(
-                    subElement => {
-                        return subElement.name === 'w:i';
+    findAndTraverseRunTimeNodes(node, nodeInformation) {
```
algomaster99, line 174: The method's purpose is to get the properties of a node. A better name would be `fetchFormattingProperties`.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
+     * @param {object} nodeInformation Contains properties and value of a node
+     * @return {object} CiceroMark Node
      */
-    getValue(elements) {
-        const variableContents = elements[1];
-        for (const property of variableContents.elements[0].elements) {
-            if (property.name === 'w:t') {
-                return property.elements[0].text;
-            }
+    construtCiceroMarkNodeJSON(nodeInformation) {
```
algomaster99, line 95: `construtCiceroMarkNodeJSON` and `constructNodes` purposes are not clear from just reading the names of the methods or even docstrings. You have even mentioned `or block CiceroMark Node` in docstrings of `constructNodes` so why was `construtCiceroMarkNodeJSON` necessary in the first place?

Perusing your code made a little clearer that former creates atomic nodes like text, softbreak, etc. While the latter creates blocks for heading paragraph. So just write their method names properly **without** relying too much on the docstrings.

algomaster99: The spelling of "construct" is wrong. I have suggested this before and I will suggest again using an IDE or a VSCode plugin for spell checkers.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,171 +87,178 @@ class OoxmlTransformer {
...
+        let obj = {};
+        if (nodeInformation.nodeType === 'softbreak') {
+            obj = {
+                $class: `${NS_PREFIX_CommonMarkModel}Softbreak`,
+            };
+        } else if (nodeInformation.nodeType === 'variable') {
+            obj = {
+                $class: `${NS_PREFIX_CiceroMarkModel}Variable`,
+                value: nodeInformation.value || 'Not provided',
```
algomaster99, line 104: If the value of the variable is `''`, you transformer will return `'Not provided`. So I think you can omit the `or` condition.

Moreover, I don't think so a variable entity would be without these attributes ever so you should omit all the `or` conditions. However, if this was possible `CiceroMark` should have defined a default _empty_ value for each of these attributes.

algomaster99 (changes requested):
Left some comments. I have more questions than changes, so please address them when you get time. Just one more question - currently, you're only dealing with nesting of bold and italic elements, right?

So I couldn't properly test it using my custom inputs, probably because there is no CLI for this right now, and I don't want to log infinitely long strings in my terminal to check the output. This is an indication that we need to integrate these transformers with CLI as soon as possible.

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: `i` is the index here for ith element. I will rename it to `index`

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: The `rootNode` here is the argument passed and not the one we get from a document which is converted via `xml-js`. The rootNode here is for paragraph heading or any other block node

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: 1. Call for the 1st element in `this.JSONXML`.
2. Call for the remaining elements present in the above array using for loop

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: > `construtCiceroMarkNodeJSON` and `constructNodes` purposes are not clear from just reading the names of the methods or even docstrings. You have even mentioned `or block CiceroMark Node` in docstrings of `constructNodes` so why was `construtCiceroMarkNodeJSON` necessary in the first place?
> 
> Perusing your code made a little clearer that former creates atomic nodes like text, softbreak, etc. While the latter creates blocks for heading paragraph. So just write their method names properly **without** relying too much on the docstrings.

`construtCiceroMarkNodeJSON`: It creates a text, softbreak or any other inline node
`constructNodes`:This creates all the nodes present in ap paragraph, or heading node.

This it the exact use of the above.

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: This was done in the case that maybe a user uses sdt in word first and then tries to convert it in a `ciceroMark`. Although it won't be a variable and I agree here so I will remove this.

K-Kumar-01:
@algomaster99 
Updated the changes.

packages/markdown-docx/src/OOXMLTransformer/index.js
algomaster99: Look at this comment too.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,200 +87,214 @@ class OoxmlTransformer {
...
-            if (property.name === 'w:t') {
-                return property.elements[0].text;
+    constructCiceroMarkNode(nodeInformation) {
+        let obj = {};
+        if (nodeInformation.nodeType === 'softbreak') {
+            obj = {
+                $class: `${NS_PREFIX_CommonMarkModel}Softbreak`,
+            };
+        } else if (nodeInformation.nodeType === 'variable') {
+            if (nodeInformation.elementType && nodeInformation.name && nodeInformation.value) {
```
algomaster99, line 102: I don't think so you need this check. There won't be any case where any of them is not there if a variable is present.

packages/markdown-docx/src/OOXMLTransformer/index.js
algomaster99: Look at this comment.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -15,22 +15,52 @@
...
     /**
-     * Gets the id of the variable
+     * Defines the JSON XML array for blocks
+     */
+    constructor() {
+        // for storing the xml tags present in a node containing its properties in array form
+        this.JSONXML = [];
+
+        // nodes of a paragraph or heading element
+        this.nodes = [];
```
algomaster99, line 34: Any reason why these are instance variables? Their content would be cleared after each block element is traversed so I think it would be better to pass such elements as parameters to functions where you need them.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,200 +87,214 @@ class OoxmlTransformer {
...
-                return {
-                    isHeading: true,
-                    level: headingLevel[headingLevel.length - 1],
+            let rootNodesLength = 1;
+            for (let nodePropertyIndex = 1; nodePropertyIndex < this.JSONXML.length; nodePropertyIndex++) {
+                let propertiesPrevious = this.JSONXML[nodePropertyIndex - 1].properties;
+                let propertiesCurrent = this.JSONXML[nodePropertyIndex].properties;
+
+                let commonPropertiesLength = 0;
+                for (let j = 0; j < Math.min(propertiesPrevious.length, propertiesCurrent.length); j++) {
```
algomaster99, line 144: Could you be cleared what `j` means? Please spend some time naming your variables.

algomaster99 (changes requested):

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: There was a case present. This is the reason i used above.

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: I was not able to properly use them in a function thats is the solo reaon i used them as above.

packages/markdown-docx/src/OOXMLTransformer/index.js
algomaster99: Is this case handled in any test case? If not, please push it.

K-Kumar-01: Actually, the test case file is deleted. I will try to recreate a similar test

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: @algomaster99 
I updated it to `constructCiceroMarkNode`. Is this also not appropriate?

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: @algomaster99 
I have pushed a commit which involves the test for this.

packages/markdown-docx/src/OOXMLTransformer/index.js
algomaster99: I have posted this comment under `let obj = {};`.

packages/markdown-docx/src/OOXMLTransformer/index.js
algomaster99:
```json
{
  "$class": "org.accordproject.commonmark.Document",
  "xmlns": "http://commonmark.org/xml/1.0",
  "nodes": [
    {
      "$class": "org.accordproject.commonmark.Heading",
      "level": "2",
      "nodes": [
        {
          "$class": "org.accordproject.commonmark.Text",
          "text": "Try TemplateMark"
        }
      ]
    },
    {
      "$class": "org.accordproject.commonmark.Paragraph",
      "nodes": [
        {
          "$class": "org.accordproject.commonmark.Text",
          "text": "Just trying "
        },
        {
          "$class": "org.accordproject.commonmark.Strong",
          "nodes": [
            {
              "$class": "org.accordproject.commonmark.Text",
              "text": "few things "
            },
            {
              "$class": "org.accordproject.commonmark.Emph",
              "nodes": [
                {
                  "$class": "org.accordproject.commonmark.Text",
                  "text": "here"
                }
              ]
            }
          ]
        },
        {
          "$class": "org.accordproject.commonmark.Softbreak"
        },
        {
          "$class": "org.accordproject.commonmark.Strong",
          "nodes": [
            {
              "$class": "org.accordproject.commonmark.Text",
              "text": "hello"
            }
          ]
        },
        {
          "$class": "org.accordproject.commonmark.Text",
          "text": " "
        },
        {
          "$class": "org.accordproject.commonmark.Strong",
          "nodes": [
            {
              "$class": "org.accordproject.commonmark.Text",
              "text": "there"
            }
          ]
        }
      ]
    }
  ]
}
```

This is the test case you pushed but I cannot see any variable entity. So your program shouldn't even enter this `if` branch.

algomaster99 (changes requested):

K-Kumar-01:
@algomaster99 
I apologize. I understood the wrong case. This test is for the `line 195`. 
As for the above-mentioned, previously it was done as I did currently. Hence, I did not change. Should I change it?

K-Kumar-01:
@algomaster99 
Made the naming changes. Let me know if there are any changes required.

packages/markdown-docx/src/OOXMLTransformer/index.js
algomaster99: To further prove my point, check t[his step](https://github.com/accordproject/markdown-transform/pull/421/checks?check_run_id=2982693335#step:6:1122) out. The `else` branch is not being covered also.

packages/markdown-docx/src/OOXMLTransformer/index.js
```diff
@@ -58,200 +87,214 @@ class OoxmlTransformer {
...
+                    if (propertiesCurrent[j] === propertiesPrevious[j]) {
+                        commonPropertiesLength++;
+                    } else {
+                        break;
```
algomaster99, line 156: This is also not covered by the coverage tool. Either your tests are weak or this is unnecessary.

algomaster99:
> As for the above-mentioned, previously it was done as I did currently. Hence, I did not change. Should I change it?

Didn't get you. Could you please be more clearer?

K-Kumar-01:
> > As for the above-mentioned, previously it was done as I did currently. Hence, I did not change. Should I change it?
> 
> Didn't get you. Could you please be more clearer?

I mean you also checked for the name, variable and type and then only defined that as variable.

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: @algomaster99 
`else` is not used currently since I am not having more inline elements which may cause this case to be called.

algomaster99:
> I mean you also checked for the name, variable and type and then only defined that as variable.

Well, you are here to improve the code, right? If the justification of your change is "Aman has done like that", you won't be able to improve it much. I wrote a broken transformer last year but take inspiration from it and please question things if they are now necessary or not.

K-Kumar-01:
> > I mean you also checked for the name, variable and type and then only defined that as variable.
> 
> Well, you are here to improve the code, right? If the justification of your change is "Aman has done like that", you won't be able to improve it much. I wrote a broken transformer last year but take inspiration from it and please question things if they are now necessary or not.

Sure, but I also have a doubt regarding this. Let's say anyone who used docx first and then he wanted to convert to ciceroMark.
Also, he/she didn't gave any `name` or `type`. Then will it be a variable?

algomaster99:
> he/she didn't give any name or type

That's only possible when someone is writing the entire OOXML on their own. Or generating it and purposefully deleting its attribute. In both cases, it's the client's fault and we cannot do anything about it.

packages/markdown-docx/src/OOXMLTransformer/index.js
algomaster99: Handle it later then. If, suppose, you don't require it later, this branch would dangle in the codebase without any purpose.

packages/markdown-docx/test/data/ciceroMark/nested.json
```diff
@@ -0,0 +1 @@
+{"$class":"org.accordproject.commonmark.Document","xmlns":"http://commonmark.org/xml/1.0","nodes":[{"$class":"org.accordproject.commonmark.Heading","level":"2","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"Try TemplateMark"}]},{"$class":"org.accordproject.commonmark.Paragraph","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"Just trying "},{"$class":"org.accordproject.commonmark.Strong","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"few things "},{"$class":"org.accordproject.commonmark.Emph","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"here"}]}]},{"$class":"org.accordproject.commonmark.Softbreak"},{"$class":"org.accordproject.commonmark.Strong","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"hello"}]},{"$class":"org.accordproject.commonmark.Text","text":" "},{"$class":"org.accordproject.commonmark.Strong","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"there"}]}]}]}
```
algomaster99, line 1: Is this test doing something different than the test you proposed in #418 ?

K-Kumar-01: @algomaster99 
it can explain why ternary operator is used in `line 195`.

algomaster99: Are you talking about [this line](https://github.com/accordproject/markdown-transform/pull/421/files#diff-f99161f5498dc7c658d94fdc1be57273146a954b6b0eb0359998d748d97cc9f1R195) (`} else if (runTimeProperties.name === 'w:b') {`)? This doesn't have a ternary operator.

K-Kumar-01: ` nodeInformation.value = runTimeNodes.elements ? runTimeNodes.elements[0].text : ' ';`
This line. I think it got shifted a bit in the latest commit.

algomaster99: Check [this step](https://github.com/accordproject/markdown-transform/runs/2979775064#step:6:1121) out. This is from the build of 00f483038c63756bf377d045a213364227dcfeeb - the commit just before you added this test case. The line which you are mentioning was already covered by the coverage tool so adding this test case is not useful for checking this line. Is it doing something else?

packages/markdown-docx/src/OOXMLTransformer/index.js
K-Kumar-01: @algomaster99 
Updated the test case it now covers above.

K-Kumar-01:
@algomaster99 
Let me know if any changes are required.

packages/markdown-docx/test/data/ciceroMark/nested.json
K-Kumar-01: The test is a special case. It does not increase coverage,  but is there to tell why ternary operator is required and not simply write the expression `runTimeNodes.elements[0].text`

algomaster99: Okay, now it makes sense.

K-Kumar-01:
@algomaster99 
Tried renaming the files.
I tried using should-parse but later gave up as it unnecessarily made the text file long.

packages/markdown-docx/test/data/ciceroMark/elements-with-different-property-at-a-common-index(0,n-1).json
```diff
@@ -0,0 +1 @@
+{"$class":"org.accordproject.commonmark.Document","xmlns":"http://commonmark.org/xml/1.0","nodes":[{"$class":"org.accordproject.commonmark.Paragraph","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"just trying "},{"$class":"org.accordproject.commonmark.Strong","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"strong"}]},{"$class":"org.accordproject.commonmark.Emph","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"italic"}]}]}]}
```
algomaster99, line 1: Sometimes I wish JSON had comments but anyway, this sounds good.

packages/markdown-docx/test/data/ciceroMark/ternary-operator-in-text.json
```diff
@@ -0,0 +1 @@
+{"$class":"org.accordproject.commonmark.Document","xmlns":"http://commonmark.org/xml/1.0","nodes":[{"$class":"org.accordproject.commonmark.Heading","level":"2","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"Try TemplateMark"}]},{"$class":"org.accordproject.commonmark.Paragraph","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"Just trying "},{"$class":"org.accordproject.commonmark.Strong","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"few things "},{"$class":"org.accordproject.commonmark.Emph","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"here"}]}]},{"$class":"org.accordproject.commonmark.Softbreak"},{"$class":"org.accordproject.commonmark.Strong","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"hello"}]},{"$class":"org.accordproject.commonmark.Text","text":" "},{"$class":"org.accordproject.commonmark.Strong","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"there"}]}]}]}
```
algomaster99, line 1: You aren't parsing the "ternary operator in text". You are parsing text nodes that only have a space. One question here - do you have an idea when MS Word interprets something as space and when as a softbreak? Shouldn't this space have been a softbreak instead? If no, could you send the markdown file you parsed this CiceroMark from?

K-Kumar-01: The markdown file is 

**strong**_italic_

Screenshot from 2021-07-05 16-51-03

K-Kumar-01: > You aren't parsing the "ternary operator in text". You are parsing text nodes that only have a space. One question here - do you have an idea when MS Word interprets something as space and when as a softbreak? Shouldn't this space have been a softbreak instead? If no, could you send the markdown file you parsed this CiceroMark from?

I am not sure but I think `xmljs` might not consider space string as elements. Otherwise, it won't throw `cannot read property 0 of undefined`(AFAIR).

algomaster99: > You aren't parsing the "ternary operator in text". You are parsing text nodes that only have a space.

Said this with respect to the file name of the test case.

K-Kumar-01: Yeah, I got that. I was just explaining the reason for the space thing which might be due to `xmljs`. I will rename the file.
I mistakenly quoteed the complete statement instead of the line
>One question here - do you have an idea when MS Word interprets something as space and when as a softbreak? Shouldn't this space have been a softbreak instead?

algomaster99: > I am not sure but I think xmljs might not consider space string as elements. Otherwise, it won't throw cannot read property 0 of undefined(AFAIR).

I don't think so XMLJS is screwing up here. It's the template mark dingus because it should have interpreted that space as softbreak.

Anyway, if there are cases like this, then it makes sense to have that ternary operator.

K-Kumar-01: @algomaster99 
Made the push. Is the current name correct?

algomaster99: Sounds good.

algomaster99 (approved):

Issue #422, opened 2021-07-02
[Build] Failing builds due to network timeouts on model loading


K-Kumar-01:
# Bug Report 🐛

Currently the build fails many times when a PR is made. A re-run is required many a times for this to pass successfully. The failure is due to [this line](https://github.com/accordproject/markdown-transform/blob/f861c6e99fb2fc1cee79e2f0ee8f09389801c8c1/packages/markdown-template/test/TemplateMarkTransformer.js#L174). 
It involves fetching model but since azure(where the models are hosted) is inconsistent, we face the error.

## Expected Behavior
The build shoud pass.

## Current Behavior
The build fails,

## Possible Solution

Passing an `offline:true` flag to options will ensure the pass.

## Steps to Reproduce

1. Make a PR.
2. The build might fail with the error that it was not able to download the model.

### Desktop
 - OS: Windows, Ubuntu
 - Browser: Chrome

## Detailed Description

Passing `offline:true` will make the build pass.

algomaster99:
@dselman @K-Kumar-01 The change is required for [more test cases](https://github.com/accordproject/markdown-transform/runs/2971814927#step:6:17180) than expected. You can see the latest build on `master` is failing.

K-Kumar-01:
@algomaster99 
I think we would need to pass the `offline: true` option in all the instances where `loadModelManger` is used.

algomaster99:
Right

K-Kumar-01:
@dselman 
`const modelManager = await ModelLoader.loadModelManager([modelFilePath]);`
Here, the third argument is the options one.
Could you tell about second argument.

I want to understand how to pass the third argument without passing the second argument/ passing the appropriate second argument. 
Thanks in advance:)

jeromesimeon:
There _is_ an issue!! @K-Kumar-01 

I still see sometimes failing tests in markdown-cli and markdown-transform but the fix in #445 should already help. I'll report when I'm done with the other packages.

PR #424, opened 2021-07-04
feat(markdown-transform): add transformation CiceroMark<->OOXML

K-Kumar-01:
Signed-off-by: K-Kumar-01 <kushalkumargupta4@gmail.com>

Adds the transformation `CiceroMark<->OOXML` in the transformation graph.

### Changes

-  Add the transformer to convert `OOXML->CiceroMark`
-  Add the transformer to convert `CiceroMark->OOXML`.

### Author Checklist


K-Kumar-01:
@dselman @algomaster99 
I have created a PR to integrate the `CiceroMark<->OOXML` in `markdown-cli`.
I was not sure how to test it.

Also, the build fails due to a test. Merging [PR](https://github.com/accordproject/markdown-transform/pull/421) will fix this.

packages/markdown-transform/lib/transform.js
```diff
@@ -221,6 +228,13 @@ const transformationGraph = {
             return t.toCiceroMark(input, options);
         },
     },
+    ooxml: {
+        docs: 'OOXML',
+        fileFormat: 'binary',
+        ciceromark_parsed: async(input) => {
+            return OoxmlTransformer.toCiceroMark(input);
```
algomaster99, line 235: This would throw an error because `toCiceroMark` is not a static method.

algomaster99 (changes requested):
Dropped the comments. Did you even try running the CLI with your changes? It was broken both ways (`CiceroMark` <-> `OOXML`) as I have elaborated in the two comments below.

algomaster99:
> I was not sure how to test it.

Try taking inspiration from `markdown-html`.

K-Kumar-01:
@algomaster99 
Had a doubt regarding `CiceroMarkJSON->OOXML`. 
The test will fail because the `OOXML` styles differ. Shall I first generate an OOXML using the transformer first and then check.

Another approach for this can be to convert the OOXML to JSON using `xmljs` and check using deep.equal().

Which approach should I go for?

packages/markdown-transform/lib/transform.js
```diff
@@ -26,6 +26,9 @@ const SlateTransformer = require('@accordproject/markdown-slate').SlateTransform
 const HtmlTransformer = require('@accordproject/markdown-html').HtmlTransformer;
 const PdfTransformer = require('@accordproject/markdown-pdf').PdfTransformer;
 const DocxTransformer = require('@accordproject/markdown-docx').DocxTransformer;
+const OoxmlTransformer = require('@accordproject/markdown-docx').OoxmlTransformer;
+const CiceroMarkToOOXMLTransfomer = require('@accordproject/markdown-docx').CiceroMarkToOOXMLTransfomer;
```
dselman, line 30: Please follow the same naming convention as the other transformers. The name of the class is the transformation, while the methods indicate what is being transformed.

dselman (changes requested):

algomaster99:
> The test will fail because the OOXML styles differ

What do you mean? You can configure the OOXML for acceptance of delivery in `markdown-transform/packages/markdown-transform/test/transform.js`. After that, you can test if it's converted correctly to `CiceroMark parsed`.

K-Kumar-01:
> > The test will fail because the OOXML styles differ
> 
> What do you mean? You can configure the OOXML for acceptance of delivery in `markdown-transform/packages/markdown-transform/test/transform.js`. After that, you can test if it's converted correctly to `CiceroMark parsed`.

I mean for the other way round, where we convert from `CiceroMark->OOXML`. 

Also, regarding Dan's changes, we would need to change the naming. Currently,
**Oomxltransformer:**  `OOXML -> CiceroMark`
**CiceroMarkToOOXMLTransformer:** `CiceroMark->OOXML`

I am thinking of keeping the former same and change the latter to `CiceroMarkTransformer`. Will it work fine?

**NOTE**
According to the docstring
`HTMLTransformer`: Converts a CiceroMark or CommonMark DOM to HTML.
`PdfTransformer`: Converts a PDF to CiceroMark DOM

Though, both the functions carry `A<->B` transformation.

dselman:
> > > The test will fail because the OOXML styles differ
> > 
> > 
> > What do you mean? You can configure the OOXML for acceptance of delivery in `markdown-transform/packages/markdown-transform/test/transform.js`. After that, you can test if it's converted correctly to `CiceroMark parsed`.
> 
> I mean for the other way round, where we convert from `CiceroMark->OOXML`.
> 
> Also, regarding Dan's changes, we would need to change the naming. Currently,
> **Oomxltransformer:** `OOXML -> CiceroMark`
> **CiceroMarkToOOXMLTransformer:** `CiceroMark->OOXML`
> 
> I am thinking of keeping the former same and change the latter to `CiceroMarkTransformer`. Will it work fine?
> 
> **NOTE**
> According to the docstring
> `HTMLTransformer`: Converts a CiceroMark or CommonMark DOM to HTML.
> `PdfTransformer`: Converts a PDF to CiceroMark DOM
> 
> Though, both the functions carry `A<->B` transformation.

Yes, we want a single class (OOXMLTransformer?) that can perform OOXML <-> CiceroMark via 2 separate methods. This is the pattern we have used for all the other transformers. 

E.g. https://github.com/accordproject/markdown-transform/blob/master/packages/markdown-html/src/HtmlTransformer.js#L40

And:

https://github.com/accordproject/markdown-transform/blob/master/packages/markdown-html/src/HtmlTransformer.js#L60

K-Kumar-01:
> > > > The test will fail because the OOXML styles differ
> > > 
> > > 
> > > What do you mean? You can configure the OOXML for acceptance of delivery in `markdown-transform/packages/markdown-transform/test/transform.js`. After that, you can test if it's converted correctly to `CiceroMark parsed`.
> > 
> > 
> > I mean for the other way round, where we convert from `CiceroMark->OOXML`.
> > Also, regarding Dan's changes, we would need to change the naming. Currently,
> > **Oomxltransformer:** `OOXML -> CiceroMark`
> > **CiceroMarkToOOXMLTransformer:** `CiceroMark->OOXML`
> > I am thinking of keeping the former same and change the latter to `CiceroMarkTransformer`. Will it work fine?
> > **NOTE**
> > According to the docstring
> > `HTMLTransformer`: Converts a CiceroMark or CommonMark DOM to HTML.
> > `PdfTransformer`: Converts a PDF to CiceroMark DOM
> > Though, both the functions carry `A<->B` transformation.
> 
> Yes, we want a single class (OOXMLTransformer?) that can perform OOXML <-> CiceroMark via 2 separate methods. This is the pattern we have used for all the other transformers.
> 
> E.g. https://github.com/accordproject/markdown-transform/blob/master/packages/markdown-html/src/HtmlTransformer.js#L40
> 
> And:
> 
> https://github.com/accordproject/markdown-transform/blob/master/packages/markdown-html/src/HtmlTransformer.js#L60

@dselman 
It would cause a single file to consist of ~450 lines(as it stands and more later). If there is no issue in that, then I will do the 
refactoring in a separate PR.

Your thoughts, @algomaster99

algomaster99:
> Your thoughts, @algomaster99

Always in for making the codebase more consistent.

dselman:
> > Your thoughts, @algomaster99
> 
> Always in for making the codebase more consistent.

If you check some of the other transforms you will see that the public (exported) class delegates to visitor classes etc. for the different transforms to keep the code modular and easier to maintain.

K-Kumar-01:
@dselman @algomaster99 
I am facing an error while I was writing the test for `markdown-cli`.
Screenshot from 2021-07-09 20-59-29
I have imported the files above similarly in the `markdown-transform` also and the tests there pass.

dselman:
> @dselman @algomaster99
> I am facing an error while I was writing the test for `markdown-cli`.
> Screenshot from 2021-07-09 20-59-29
> I have imported the files above similarly in the `markdown-transform` also and the tests there pass.

My guess is that your input OOXML is being used as a file path. Note the ENAMETOOLONG error.

K-Kumar-01:
@dselman @algomaster99 
Requesting your review

packages/markdown-transform/lib/transform.js
```diff
@@ -165,6 +166,10 @@ const transformationGraph = {
         pdfmake: (input, parameters, options) => {
             return PdfTransformer.ciceroMarkToPdfMake(input, options);
         },
+        ooxml: (input) => {
+            const t = new OOXMLTransformer();
+            return t.toOOXML(input).ooxml;
```
dselman, line 171: Do we need the `ooxml` child attribute here? The other transformations just return the output.

packages/markdown-transform/lib/transform.js
```diff
@@ -221,6 +226,14 @@ const transformationGraph = {
             return t.toCiceroMark(input, options);
         },
     },
+    ooxml: {
```
dselman, line 229: Don't you also need to register the ciceromark_parsed -> ooxml transform?

dselman (changes requested):
Getting close! Couple of questions...

packages/markdown-transform/lib/transform.js
K-Kumar-01: @dselman 
Yes, actually `OOXMLTransfomer` returns `{ooxml, variableFrequency}`.

packages/markdown-transform/lib/transform.js
K-Kumar-01: The transformer is converting  both `ooxml<->Ciceromark `

packages/markdown-transform/lib/transform.js
dselman: What is variable frequency and why does the caller need it?

packages/markdown-transform/lib/transform.js
dselman: My point is it looks like you only registered the ciceromark_parsed -> ooxml transformation. You will need to add a similar block to do the reverse, under a new `ooxml` block.

dselman: Please check that you can transform from a markdown file to ooxml and the reverse by running the CLI.

packages/markdown-transform/lib/transform.js
K-Kumar-01: It returns the count of different variables present in a template. 
We need it because there is a requirement to provide a unique identifier to `rich text input` in MS-WORD which is generated using this.

packages/markdown-transform/lib/transform.js
K-Kumar-01: @dselman 
AFAIR, I was able to convert both transformations `ooxml<->ciceromark_parsed` when I pushed the commit. However, I will recheck and let you know ASAP.

K-Kumar-01: @dselman 
Apologies for the delay. I have checked and it converts `CiceroMark_parsed <-> OOXML`. So, this feature is as we expected.

packages/markdown-cli/test/cli.js
```diff
@@ -148,3 +148,17 @@ describe('markdown-cli (docx)', () => {
...
+    const inputOOXMLFile = path.resolve(__dirname, 'data/acceptance', 'partial_acceptance.xml');
+    const expectedCiceroMarkParsed = JSON.parse(fs.readFileSync(path.resolve(__dirname, 'data/acceptance', 'partial_acceptance.json'), 'utf8'));
+
+    describe('#parse', () => {
+        it('should generate a ciceromark_parsed from ooxml', async () => {
+            const result = await Commands.transform(inputOOXMLFile, 'ooxml', [], 'ciceromark_parsed', null, {}, {});
+            JSON.parse(result).should.deep.equal(expectedCiceroMarkParsed);
+        });
+    });
+});
```
algomaster99, line 163: You can probably check for roundtrip by passing the `roundtrip: true` option. [Check example](https://github.com/accordproject/markdown-transform/pull/424/files#diff-f157946de8148b534c64faaec27c403cfe2a77b108cb617830f91982c17b9fb0R133).

packages/markdown-transform/lib/transform.js
algomaster99: > What is variable frequency and why does the caller need it?

MS Word would require it for rendering all the variables.

packages/markdown-transform/test/transform.js
```diff
@@ -312,3 +319,10 @@ describe('#formatDescriptor', () => {
         (() => formatDescriptor('foobar')).should.throw('Unknown format foobar');
     });
 });
+
+describe('#ooxml', () => {
+    it('ooxml->ciceromark_parsed', async () => {
```
algomaster99, line 324: Spaces before and after `->`.

packages/markdown-transform/test/data/acceptance/partial_acceptance.json
```diff
@@ -0,0 +1 @@
...org.accordproject.commonmark.Text","text":"Acceptance Criteria."}]},{"$class":"org.accordproject.commonmark.Paragraph","nodes":[{"$class":"org.accordproject.commonmark.Text","text":"The \"Acceptance Criteria\" are the specifications the "},{"$class":"org.accordproject.ciceromark.Variable","value":"\"Widgets\"","name":"deliverable","elementType":"String"},{"$class":"org.accordproject.commonmark.Softbreak"},{"$class":"org.accordproject.commonmark.Text","text":"must meet for the "},{"$class":"org.accordproject.ciceromark.Variable","value":"\"Party A\"","name":"shipper","elementType":"org.accordproject.organization.Organization"},{"$class":"org.accordproject.commonmark.Text","text":" to comply with its requirements and"},{"$class":"org.accordproject.commonmark.Softbreak"},{"$class":"org.accordproject.commonmark.Text","text":"obligations under this agreement, detailed in "},{"$class":"org.accordproject.ciceromark.Variable","value":"\"Attachment X\"","name":"attachment","elementType":"String"},{"$class":"org.accordproject.commonmark.Text","text":", attached"},{"$class":"org.accordproject.commonmark.Softbreak"},{"$class":"org.accordproject.commonmark.Text","text":"to this agreement."}]}]}
```
algomaster99, line 1: What all have you removed from the original `acceptance-of-delivery`?

algomaster99 (changes requested):

packages/markdown-transform/test/data/acceptance/partial_acceptance.json
K-Kumar-01: @algomaster99 
I have not removed the original `acceptance-of-delivery`. I just added one more named as `partial-acceptance`

algomaster99: I meant what all entities have you removed to create parital_acceptance.

K-Kumar-01: `CLuaseDefinition`. This was the only entity causing hinderance.

algomaster99:
@K-Kumar-01 Did you get time to look at the changes I suggested?

K-Kumar-01:
> @K-Kumar-01 Did you get time to look at the changes I suggested?

@algomaster99 
I am sorry it slipped my mind. I will push the changes surely by today.

algomaster99 (approved):
Looks good to me. The tests are failing though, and the reason is not #422 .

K-Kumar-01:
> Looks good to me. The tests are failing though, and the reason is not #422 .

I have the reason. I deleted the json for the wrong package, mistakenly. 😅
I didn't re-run the test as it was passing.
I will update the PR ASAP and then we can merge it 😀

packages/markdown-cli/test/cli.js
```diff
@@ -148,3 +148,16 @@ describe('markdown-cli (docx)', () => {
...
+
+describe('markdown-cli (ooxml)', () => {
+    // Partial Acceptance test
+    const inputOOXMLFile = path.resolve(__dirname, 'data/acceptance', 'partial_acceptance.xml');
+
+    describe('#parse', () => {
+        it('should roundtrip ciceromark_parsed <-> ooxml', async () => {
+            const result = await Commands.transform(inputOOXMLFile, 'ooxml', [], 'ciceromark_parsed', null, {}, { roundtrip: true });
+            result.should.equal(fs.readFileSync(inputOOXMLFile, 'utf-8'));
+        });
```
algomaster99, line 160: I think you would also need to check for going from cicero mark to OOXML. And for that, you would need to create a JSON file partial_acceptance.json.

algomaster99 (changes requested):
I have another general suggestion about the naming of the test file and I know it is coming quite in this review process. Partial acceptance sounds like that your test is _paritally_ correct. I think `omitted-acceptance-of-delivery` would be a better name.

@dselman I think @K-Kumar-01 is done with this. Could you look at it again?

packages/markdown-cli/test/cli.js
K-Kumar-01: @algomaster99 Will update the filenames and tests ASAP.

packages/markdown-transform/lib/transform.js
dselman: Can you explain why, please? How does variable frequency help?

packages/markdown-cli/test/cli.js
K-Kumar-01: > I think you would also need to check for going from cicero mark to OOXML. And for that, you would need to create a JSON file partial_acceptance.json.

@algomaster99 
Regarding this, we already check for the conversion test so why do we do this one. `roundtrip: true` also pass so I am not able to understand the need to add this.
In `transform.js`, we have already done the test for both.

packages/markdown-transform/lib/transform.js
algomaster99: Sure!

In MS Word, `org.accordproject.ciceromark.Variable` is converted to a `w:sdt` XML tag. This XML tag represents content controls in MS Word. A content control has many attributes but we use only two of them - `title` and `tag` (think of them as analogous to `id` and `class` in HTML, respectively).

The `title` attribute is compulsory and **has to be unique** but since there can be many instances of a variable in a clause, we cannot assign each of them the same title. For example, all four instances of `Shipper` in `acceptance-of-delivery` cannot have the same title `Shipper` so I decided to maintain a counter object which would keep a count of the occurrences of the variable and assign the title as - `Shipper1`, `Shipper2`, `Shipper3`, and `Shipper4`. We are also exporting the `counter` variable as MS Word will use it to attach listeners [here](https://github.com/accordproject/cicero-word-add-in/blob/master/src/components/TemplateLibrary/index.js#L146).

In retrospect, I think exporting counter variable _may not_ be required. I will try to look into it and discuss this with you and @K-Kumar-01 in the next meeting.

@K-Kumar-01 , for now, maybe we can revert exporting of `counter` object and do it when we really find that necessary.

K-Kumar-01: @algomaster99 I dont like the idea of not exporting the `counter`. Reason being i do believe that we would need it a later point, basically when we use it with add-in. On what basis so you think that we might not require the variable? Could you please explain that?

packages/markdown-cli/test/cli.js
algomaster99: > Regarding this, we already check for the conversion test so why do we do this one. roundtrip: true also pass so I am not able to understand the need to add this.

We aren't checking for equality for ciceromark_parsed in between so that's why it's needed.

> In transform.js, we have already done the test for both.

Yeah, I am aware. I even pointed this out in the meeting. With that logic, I think we can completely remove any testing related to this PR from `markdown-cli`.

packages/markdown-transform/lib/transform.js
algomaster99: Cicero Word add-in uses the counter variable to attach listeners. Instead of the for-loop, we can fetch all the content controls using [`body.contentControls`](https://docs.microsoft.com/en-us/javascript/api/word/word.body?view=word-js-preview#contentControls) and then attach the listener one-by-one. However, there are two questions to ponder over first:
1. What happens if the event listener is attached again to a content control which already had an event listener attached?
2. What happens when the content control with a non-unique title is inserted?

dselman: It should be possible to generate a unique id for a variable based on its position in the tree. E..g using the path to the node. Alternatively we might be able to assign a UUID further up in the transformation stack. Let's discuss some ideas.

algomaster99: Generating a unique idea is not the problem but getting back to attach event listeners is. Moreover, we want to keep this ID as close as possible to the actual variable name as they are means to inform the client which variable they are interacting with.

cc

`Shipper1` is the title and if someone edits, it would mean that they want to modify the shipper variable.

algomaster99: I am looking for ways to make us _not require_ to export the `counter` variable. But we shall use it for OOXML generation as we want to make the content control unique.

algomaster99: @K-Kumar-01 I think you can remove exporting of `counter`. I can demonstrate how we can do without it in the next meeting.

K-Kumar-01: @algomaster99 
That would need refactoring first then. I will create a separate PR that will deal with the above.
Once that is done, I will merge it.
Since, currently only windows are available on my pc it might take some time to close these PRs.

PS: HDD corruption made Linux vanish 😢. Will reinstall it.

algomaster99: Sure. No problem.

K-Kumar-01:
@dselman @algomaster99 
Screenshot from 2021-07-23 17-24-52
Screenshot from 2021-07-23 17-25-36
Screenshot from 2021-07-23 17-25-40
Screenshot from 2021-07-23 17-25-41

I am not understanding why the test fails.
Both the string seems to be identical to me.
Can any one you help me here. After this is resolved, this PR will get ready to be merged.

algomaster99:
@K-Kumar-01 It's difficult to find out this way. You can use diff editors or just run `diff` over them to understand the difference.

algomaster99 (changes requested):
One minor change and then we can merge.

Also, you haven't merged the latest changes correctly and that's why changes related to thematic break, etc are appearing in this PR.

packages/markdown-transform/test/transform.js
```diff
@@ -55,6 +55,8 @@ const acceptanceCiceroMarkUnquoted = JSON.parse(fs.readFileSync(path.resolve(__d
 const acceptanceSlate = JSON.parse(fs.readFileSync(path.resolve(__dirname, 'data/acceptance', 'slate.json'), 'utf8'));
 const acceptancePlainText = normalizeNLs(fs.readFileSync(path.resolve(__dirname, 'data/acceptance', 'sample.txt'), 'utf8'));
 const acceptanceHtml = normalizeNLs(fs.readFileSync(path.resolve(__dirname, 'data/acceptance', 'sample.html'), 'utf8'));
+const partialAcceptanceXML = fs.readFileSync(path.resolve(__dirname, 'data/acceptance', 'omitted-acceptance-of-delivery.xml'), 'utf8');
+const partialAcceptanceCiceroMarkParsed = JSON.parse(fs.readFileSync(path.resolve(__dirname, 'data/acceptance', 'omitted-acceptance-of-delivery.json'), 'utf8'));
```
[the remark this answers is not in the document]
K-Kumar-01: @algomaster99 
Done

algomaster99 (approved):
@dselman I think this is done. Could you have a final look and if you think alike, go ahead with **squash and merge**?

PR #436, opened 2021-08-05
feat(markdown-docx): add link transformer - #397

K-Kumar-01:

Add the transformation logic for `link`.

### Changes

- transformation logic: OOXML<->CiceroMark
- rules: link and link property
- tests: link and its variants

### Flags

- <ONE> Syntactical differences may occur. Example case: `strong->hyperlink->text` will always get converted to `hyperlink->strong->text`

### Author Checklist


algomaster99 (changes requested):
I tried to render the generated OOXML in MS Word but I didn't see any of the links behave like actual links. I am assuming they are supposed to redirect me somewhere when I click on them. What should be the actual behaviour?

link

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -268,6 +280,14 @@ class ToCiceroMarkVisitor {
...
                         constructedNode,
                     ];
+                } else if (commonPropertiesLength === 3) {
+                    const subNodeLength = rootNode.nodes[rootNodesLength - 1].nodes.length;
+                    const deepSubNodeLength = rootNode.nodes[rootNodesLength - 1].nodes[subNodeLength - 1].nodes.length;
+                    rootNode.nodes[rootNodesLength - 1].nodes[subNodeLength - 1].nodes[deepSubNodeLength - 1].nodes = [
+                        ...rootNode.nodes[rootNodesLength - 1].nodes[subNodeLength - 1].nodes[deepSubNodeLength - 1]
+                            .nodes,
+                        constructedNode,
+                    ];
```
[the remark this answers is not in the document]
K-Kumar-01: @algomaster99 
I intend to refactor many things after this PR gets merged. Many things can be reused.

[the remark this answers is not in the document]
K-Kumar-01: @algomaster99 For this function, I am unsure.

K-Kumar-01:
> I tried to render the generated OOXML in MS Word but I didn't see any of the links behave like actual links. I am assuming they are supposed to redirect me somewhere when I click on them. What should be the actual behaviour?
> 
> link

@algomaster99 
I have generated the OOXML and on click it opens a `tab` in my Chrome with the `url` specified in the link.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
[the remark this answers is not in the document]
algomaster99: > I intend to refactor many things after this PR gets merged. Many things can be reused.

Sounds great!

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -47,6 +49,8 @@ class ToOOXMLVisitor {
         this.counter = {};
         // OOXML tags for a given block node(heading, paragraph, etc.)
         this.tags = [];
+        // Relationship tags for links in a document
+        this.relationShips = [];
```
algomaster99, line 53: `relationship` is a single word. No need to capitalise `S`.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -86,21 +94,40 @@ class ToOOXMLVisitor {
...
                     }
                     propertyTag = TEXT_STYLES_RULE(propertyTag);
 
                     let textValueTag = TEXT_RULE(subNode.text);
 
                     let tag = TEXT_WRAPPER_RULE(propertyTag, textValueTag);
+                    if (isLinkPropertyPresent) {
+                        // some rels are already present
+                        // To avoid overwrite ids start our
+                        // rels from 10
```
algomaster99, line 127: What do you mean by this? Could you elaborate?

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -72,11 +76,15 @@ class ToOOXMLVisitor {
...
                     let propertyTag = '';
+                    let isLinkPropertyPresent = false;
                     for (let property of properties) {
                         if (property === TRANSFORMED_NODES.emphasize) {
                             propertyTag += EMPHASIS_RULE();
                         } else if (property === TRANSFORMED_NODES.strong) {
                             propertyTag += STRONG_RULE();
+                        } else if (property === TRANSFORMED_NODES.link) {
+                            isLinkPropertyPresent = true;
+                            propertyTag+= LINK_PROPERTY_RULE();
```
algomaster99, line 87: Your formatter didn't add space between `propertyTag` and `+=`.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -446,9 +477,18 @@ class ToCiceroMarkVisitor {
...
             } else if (subNode.name === 'w:r') {
                 let nodeInformation = { properties: [], value: '' };
-                this.fetchFormattingProperties(subNode, nodeInformation);
+                if (parent === TRANSFORMED_NODES.link) {
+                    nodeInformation.linkId = id;
+                }
+                this.fetchFormattingProperties(
+                    subNode,
+                    nodeInformation,
+                    parent === TRANSFORMED_NODES.link ? TRANSFORMED_NODES.link : TRANSFORMED_NODES.paragraph
```
algomaster99, line 490: Are there only two types of parents possible? This looks quite incomprehensible and an attempt to hardcode the logic. If it's really required, maybe you can few lines of comments elucidating this.

algomaster99 (changes requested):

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
If you see the below:
```xml
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
          <Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/numbering" Target="numbering.xml"/>
          <Relationship Id="rId2" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/styles" Target="styles.xml"/>

        </Relationships>
```

Here some rels are already used basically for styling, themes, or settings, etc.
Hence, to avoid conflict I start with 10.
Actually, it would be `11`. I will correct this typo

packages/markdown-docx/src/ToCiceroMarkVisitor.js
K-Kumar-01: @algomaster99 
Nope, many parents can be possible, but when we call this we only need to check whether this function is called by Link or not (particularly these lines). So I have passed `link` if true else `paragraph`.

The main reason for not naming it as `calledByLink` is because we use similar things in `clause` also.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
Will update this.

K-Kumar-01:
@algomaster99 
Made some changes. Let me know if any more changes are required.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -337,11 +360,17 @@ class ToCiceroMarkVisitor {
...
-     * @param {object} node Node object to be traversed
+     * @param {object} node   Node object to be traversed
      * @param {string} parent Parent node name
+     * @param {syring} id     Relation Id for link in OOXML
      * @returns {*} GeneratedNode if parent is of type clause else none
      */
-    traverseElements(node, parent = '') {
+    traverseElements(node, parent = TRANSFORMED_NODES.paragraph, id = undefined) {
+        /**
+         * The parent argument is useful in cases where parent is a clause or link.
```
algomaster99, line 370: A good way to explain something is to write what would have happened if this change was not there.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -86,21 +94,40 @@ class ToOOXMLVisitor {
                     let textValueTag = TEXT_RULE(subNode.text);
 
                     let tag = TEXT_WRAPPER_RULE(propertyTag, textValueTag);
+
+                    if (isLinkPropertyPresent) {
+                        // some rels are already present
```
algomaster99, line 99: You can just write 'relationships'. I see no need to abbreviate it.

```js
// Two elationships for numbering and style are already present
// and since we need to accommodate for link styles as well, we need a unique ID
// to represent them. Hence, 2 is added to offset the enumeration of `rId`.
```

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -86,21 +94,40 @@ class ToOOXMLVisitor {
...
+                        // To avoid overwrite ids start our
+                        // rels from 10
+                        let relationShipId = 'rId' + (this.relationships.length + 10).toString();
```
algomaster99, line 102: 2 is sufficient, I think. We only have two relationship tags as of now.

algomaster99: `relationshipId`. Please change it everywhere.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -86,21 +94,40 @@ class ToOOXMLVisitor {
...
                     let textValueTag = TEXT_RULE(subNode.text);
 
                     let tag = TEXT_WRAPPER_RULE(propertyTag, textValueTag);
+                    if (isLinkPropertyPresent) {
+                        // some inbuilt rels are already present for settings, theme, etc.
+                        // To avoid overwrite ids or conflicts start our
+                        // rels from 11
+                        let relationShipId = 'rId' + (this.relationships.length + 10).toString();
+                        tag = LINK_RULE(tag, relationShipId);
+                    }
```
algomaster99, line 130: Why is there a repetition of the same code? Please follow DRY as much as possible.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -224,6 +251,14 @@ class ToOOXMLVisitor {
                             this.globalOOXML += ooxml;
                             this.tags = [];
                         } else {
+                            if (this.getClass(subNode) === TRANSFORMED_NODES.link) {
+                                const relationshipTag = `<Relationship Id="rId${
+                                    this.relationships.length + 11
+                                }" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="${
+                                    subNode.destination
+                                }" TargetMode="External"/>`;
+                                this.relationships = [...this.relationships, relationshipTag];
```
algomaster99, line 260: I think it would be better to just collect `subNode.destination` here. And then you can pass that array to `RELATIONSHIP_SPEC`.

algomaster99 (changes requested):

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
We would need the `id` also because that defines what connects what.
If we pass it to `RELATIONSHIP_SPEC`, then we would need to construct the tags there eventually. I don't understand the reason for passing it like this.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
I decided to refactor this in a new PR. But will do in the same for now as it is related to `link` only.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 I am unsure about 2. I would at least keep it 6/7 as word default enters 5-6 relationship Ids.

K-Kumar-01: @algomaster99 Currenlty, I went with 2 in the latest commit.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -59,6 +63,46 @@ class ToOOXMLVisitor {
...
 
+    /**
+     * Generates the OOXML for text and code ciceromark nodes.
+     *
+     * @param {string} value           Text value of the node
+     * @param {Array}  nodeProperties  Properties of the node
+     * @param {string} defaultProperty Default property of the node
+     * @returns {striing} Generated OOXML
+     */
+    generateTextOrCodeOOXML(value, nodeProperties, defaultProperty = '') {
```
algomaster99, line 74: I don't think so defaultProperty should be an optional parameter. Do you have any case where `''` is a valid argument? Moreover, its name is also not descriptive. I think it should be renamed to `stylisticProperties`.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -59,6 +63,46 @@ class ToOOXMLVisitor {
...
+        let propertyTag = defaultProperty;
```
algomaster99, line 75: What is the point of this? You can directly use the argument's name.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -224,6 +242,14 @@ class ToOOXMLVisitor {
...
                             this.tags = [];
                         } else {
+                            if (this.getClass(subNode) === TRANSFORMED_NODES.link) {
+                                const relationshipTag = `<Relationship Id="rId${
+                                    this.relationships.length + 3
+                                }" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="${
+                                    subNode.destination
+                                }" TargetMode="External"/>`;
+                                this.relationships = [...this.relationships, relationshipTag];
+                            }
```
algomaster99, line 252: Convert the links into tags in the file which contains the `LINK_SPEC`.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -224,6 +242,14 @@ class ToOOXMLVisitor {
                             this.globalOOXML += ooxml;
                             this.tags = [];
                         } else {
+                            if (this.getClass(subNode) === TRANSFORMED_NODES.link) {
+                                const relationshipTag = `<Relationship Id="rId${
+                                    this.relationships.length + 3
```
algomaster99, line 247: Why is `3` used here? You used `2` above. It is better to hardcode that offset and then reuse it everywhere.

packages/markdown-docx/test/data/ciceroMark/link.json
```diff
@@ -0,0 +1,36 @@
+{
```
algomaster99, line 1: All of your other test cases are uglified. Any specific reason to prettify this?

I would prefer it prettified, to be honest, but I thought you deliberately wanted to keep all the uglified when you wrote the first test case.

algomaster99 (changes requested):

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
In `code` only there is a requirement of `defaultProperty`.For rest of the text nodes, its empty.
I will change some arguments  to make it more understandable.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
Didn't understand what you mean.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
At one moment, the length is zero. After that the relationship tag is append. This makes the length increase by 1. Hence, the difference.

packages/markdown-docx/test/data/ciceroMark/link.json
K-Kumar-01: @algomaster99 
I might have missed this thing.
I can prettify the test cases in a different PR.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
algomaster99: I am questioning the assignment of defaultProperty argument to `propertyTag`. You can just use defaultProperty instead of propertyTag.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
algomaster99: So it would be better to use `RELATIONSHIP_OFFSET +1`.

packages/markdown-docx/src/constants.js
```diff
@@ -25,13 +25,16 @@ const TRANSFORMED_NODES = {
...
     strong: `${NS_PREFIX_CommonMarkModel}Strong`,
     text: `${NS_PREFIX_CommonMarkModel}Text`,
     thematicBreak: `${NS_PREFIX_CommonMarkModel}ThematicBreak`,
     variable: `${NS_PREFIX_CiceroMarkModel}Variable`,
-    clause:`${NS_PREFIX_CiceroMarkModel}Clause`
+    clause: `${NS_PREFIX_CiceroMarkModel}Clause`,
 };
 
-module.exports = { TRANSFORMED_NODES };
+const RELATIONSHIP_OFFSET = 2;
```
algomaster99, line 38: Please add comments because I don't think so anyone can infer what this offset is for. The best idea would be to keep this in the `helpers.js` file only but it can't go there semantically.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -59,6 +63,49 @@ class ToOOXMLVisitor {
...
+        }
+
+        let textValueTag = TEXT_RULE(value);
+
+        let tag = TEXT_WRAPPER_RULE(propertyTag, textValueTag);
+
+        if (isLinkPropertyPresent) {
+            // Two relationships for numbering and style are already present
+            // and since we need to accommodate for link styles as well, we need a unique ID
+            // to represent them. Hence, 2 is added to offset the enumeration of `rId`.
```
algomaster99, line 101: This comment would go in that constants file now. Ensure that you link the filename too.

algomaster99 (changes requested):
Final changes and then we can merge.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -214,6 +217,15 @@ class ToCiceroMarkVisitor {
                 $class: nodeInformation.properties[nodePropertyIndex],
                 nodes: [ciceroMarkNode],
             };
+            if (nodeInformation.properties[nodePropertyIndex] === TRANSFORMED_NODES.link) {
+                ciceroMarkNode.title = '';
```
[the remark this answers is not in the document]
K-Kumar-01: @algomaster99 
That would add a `title` key in each node which we do not want. SO I guess this is not possible.

packages/markdown-docx/src/constants.js
K-Kumar-01: @algomaster99 
I would prefer keeping it in constants and adding suitable comments here itself.

packages/markdown-docx/src/ToOOXMLVisitor/helpers.js
```diff
@@ -211,17 +227,25 @@ function wrapAroundDefaultDocxTags(ooxml) {
...
+      ${LINK_STYLE_SPEC}
     </w:styles>
   </pkg:xmlData>
   </pkg:part>
     `;
 
+    let relationshipOOXML = '';
+    relationships.forEach(
+        ({ id, destination }) =>
+            (relationshipOOXML += `<Relationship Id="rId${id}" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="${destination}" TargetMode="External"/>`)
```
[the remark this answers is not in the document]
K-Kumar-01: @algomaster99 
The above is a single tag, not multiple tags.So should I break it ?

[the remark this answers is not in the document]
algomaster99: It has multiple attributes. You can break on each attribute like you do in react.

But it might be syntactically incorrect because word is weird. If you are in a rush, you can let this change be as is.

algomaster99 (approved):

PR #441, opened 2021-08-10
feat(markdown-docx): add optional transformer - #397

K-Kumar-01:

Add the logic to transform `Optional` nodes from `CiceroMark<->OOXML`.

### Changes

- transformation logic:OOXML<->CiceroMark
- tests for optional
- update tests for other(xml changes in variable and clause)

### Author Checklist


dselman (commented):
I'm finding the code hard to follow, as you add more special cases. In general terms I would structure the code as a transformation from treeA -> treeB with a switch statement to handle different nodes in treeA, converting them to nodes in treeB. In particular I found the part where you compare lengths of arrays to hard coded values confusing.

packages/markdown-cli/test/data/acceptance/omitted-acceptance-of-delivery.xml
```diff
@@ -71,7 +71,7 @@
             <w:sz w:val="24"/>
           </w:rPr>
           <w:alias w:val="Shipper1 | org.accordproject.organization.Organization"/>
-          <w:tag w:val="shipper"/>
+          <w:tag w:val="org.accordproject.ciceromark.Variable----shipper"/>
```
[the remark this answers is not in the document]
K-Kumar-01: `w:alias`: the name which is displayed by word on hovering on content control.
`w:tag`: Bind the office JS API handles for elements with similar tags.

>  Could we make them consistent. E.g.org.acme.Thing/name ?

I do not understand what exactly you mean here. I have introduced the first thing to distinguish the type of content control as we discussed in the meeting.

K-Kumar-01:
@algomaster99 
Requesting your reviews also.

K-Kumar-01:
@dselman @algomaster99 
I have completed the transformation for the `formula` in another branch. Since the branch is extension of this one, we can merge that once this PR gets merged.

algomaster99 (changes requested):
First iteration of review done. Look at the comments.

Also, add a test case when `hasSome` is `false`. You will need to see how `nodes` behave in that case. Currently, you have made `nodes` = `whenSome`.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -87,19 +90,16 @@ class ToCiceroMarkVisitor {
...
     getNodeType(properties) {
         let nodeType = TRANSFORMED_NODES.variable;
         for (const property of properties) {
-            if (property.name === 'w15:color') {
-                // eg. "Shipper1 | org.accordproject.organization.Organization"
-                if (property.attributes['w:val'] === '99CCFF') {
-                    nodeType = TRANSFORMED_NODES.clause;
-                }
+            if (property.name === 'w:tag') {
+                return property.attributes['w:val'].split('----')[0];
```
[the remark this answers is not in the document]
K-Kumar-01: @algomaster99 what is meant by soft code this separator?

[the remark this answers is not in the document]
algomaster99: Create a constant for it so you don't need to hardcode `|` wherever you want to split text.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -51,6 +52,23 @@ class ToOOXMLVisitor {
...
+        // in optional this means if hasSome is true then whenNone are these nodes
+        // if hasSome is false then whenSome nodes are the corresponding required nodes
+        // this is done to add vanish property
+        this.isTraversingFalseCondition = false;
+
+        // Whether the nodes traversed are present in else condition
+        // done to add specific font property
+        // font property+ vanish = whenNone, else nodes when converting from
+        // ooxml->ciceromark. simple vanish means whenNone, if nodes
+        this.isTraversingElseCondition = false;
```
algomaster99, line 71: It's quite dangerous to have class variables like these and then use a method to flip their values. I would recommend you to make this function local to the block where you check if the transformed node is `Optional`.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -87,6 +105,12 @@ class ToOOXMLVisitor {
                 propertyTag += LINK_PROPERTY_RULE();
             }
         }
+        if (this.isTraversingFalseCondition) {
+            propertyTag += '<w:vanish/>';
+        }
+        if (this.isTraversingElseCondition) {
+            propertyTag += '<w:rFonts w:ascii="Baskerville Old Face" w:hAnsi="Baskerville Old Face"/>';
```
algomaster99, line 112: This should be in the rules file. You have put most properties and nodes in `rules.js` so why have you made an exception here?

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -102,23 +126,55 @@ class ToOOXMLVisitor {
...
      * Traverses CiceroMark nodes in a DFS approach
      *
-     * @param {object} node       CiceroMark Node
-     * @param {array}  properties Properties to be applied on current node
+     * @param {object}  node                           CiceroMark Node
+     * @param {array}   properties                     Properties to be applied on current node
+     * @param {boolean} isConditionalOrOptionalPresent True if the dfs traversal contains conditional or optional else false
      */
-    traverseNodes(node, properties = []) {
+    traverseNodes(node, properties = [], isConditionalOrOptionalPresent = '') {
```
algomaster99, line 159: The docstrings say it should be boolean however you have assigned it an empty string.

packages/markdown-docx/src/constants.js
```diff
@@ -41,4 +42,10 @@ const TRANSFORMED_NODES = {
 // Used in './ToOOXMLVisitor/index.js'
 const RELATIONSHIP_OFFSET = 2;
 
-module.exports = { TRANSFORMED_NODES, RELATIONSHIP_OFFSET };
+// Used to separate entities in w:tag and w:alias
+// example <w:tag w:val="org.accordproject.ciceromark.Variable | shipper"/>
+// Then use the separator to extract those entities while transforming
+// from ooxml to ciceromark.
+const SEPARATOR = ' | ';
```
algomaster99, line 49: Figure out which separator to use as discussed in the meeting.

packages/markdown-docx/src/ToOOXMLVisitor/rules.js
```diff
@@ -121,26 +122,28 @@ const STRONG_RULE = () => {
...
  *
- * @param {string} title Title of the variable. Eg. receiver-1, shipper-1
- * @param {string} tag   Name of the variable. Eg. receiver, shipper
- * @param {string} value Value of the variable
- * @param {string} type  Type of the variable - Long, Double, etc.
+ * @param {string}  title  Title of the variable. Eg. receiver-1, shipper-1
+ * @param {string}  tag    Name of the variable. Eg. receiver, shipper
+ * @param {string}  value  Value of the variable
+ * @param {string}  type   Type of the variable - Long, Double, etc.
+ * @param {boolean} vanish Should vanish property be present
```
algomaster99, line 129: Add a link to the `vanish` property in the doc strings.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -51,6 +52,23 @@ class ToOOXMLVisitor {
         this.tags = [];
         // Relationship tags for links in a document
         this.relationships = [];
+        // Contains the tags for optional or conditional nodes.
+        // Optional/conditional themselves can act contain multiple inline nodes.
+        // However, they are not block elements and are present inside
+        // paragraph or clause nodes so an extra array is needed to store their info.
```
algomaster99, line 58: Even a paragraph contains multiple nodes but we don't maintain a different array for it. You're using a DFS approach which should technically allow you to traverse each optional node _first_ and then it would move forward. I will try to write a pseudo code for it and I hope it gives you an idea.

```
traverseNodes() {
   if OPTIONAL { 
      ooxmlToBeDisplayed = traverseNodes(whenSome) // ofcouse this depends upon `hasSome` so you would need a check for it.
      ooxmlToBeHidden = travserseNodes(whenNode)
  }
}
```

I didn't bother with `relationships` because I knew you had to store them in a different part of OOXML (some node above `w:document`) so you needed to store them until your transformation is over.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -240,6 +294,35 @@ class ToOOXMLVisitor {
...
+                            this.isTraversingFalseCondition = false;
+                            this.isTraversingElseCondition = false;
+                            this.conditionalOrOptionalTags = [];
+
+                            const tag = subNode.name;
+                            const type = subNode.elementType;
+                            this.createOrUpdateCounter(tag, type);
+                            const title = `${tag.toUpperCase()[0]}${tag.substring(1)}${this.counter[tag].count}`;
+                            let optionalTag = OPTIONAL_RULE(title, tag, ooxml, type);
+                            this.tags = [...this.tags, optionalTag];
```
algomaster99, line 325: It's so involving that I think you probably need to abstract it into a different function after considering my strongly suggested recommendation, written above, for refactoring.

Also, the way you have assigned the boolean values are redundant.
**BAD**
```js
if (!subNode.hasSome) {
    this.isTraversingFalseCondition = false;
} else {
    this.isTraversingFalseCondition = true;
}
```
**OK**
```js
this.isTraversingFalseCondition = subNode.hasSome;
```

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -240,6 +294,35 @@ class ToOOXMLVisitor {
...
+                            if (!subNode.hasSome) {
+                                this.isTraversingFalseCondition = false;
+                            } else {
+                                this.isTraversingFalseCondition = true;
+                            }
+                            this.isTraversingElseCondition = true;
+                            // traverse whenNone properties now
+                            this.traverseNodes(subNode.whenNone, properties, true);
+                            let ooxml = '';
+                            for (let tag of this.conditionalOrOptionalTags) {
```
algomaster99, line 313: It's better to use `const` inside for-loops because you don't want side-effects.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
algomaster99: I don't think so this parameter should be even present here given the function name. `traverseNodes` is such a generic name and it makes sense that it would have `node` and `properties` but `isConditonalOrOptional` just takes away is "genericity".

I hope this is resolved by the pseudocode I suggested.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
algomaster99: I think this would require a refactoring of `traverseNodes` as it doesn't return anything unlike `toCiceroMark`. Maybe you need to keep that on priority because the current code is getting out of hand in terms of readability.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -270,24 +286,114 @@ class ToCiceroMarkVisitor {
...
+                            rootNode.nodes[rootNodesLength - 1].nodes = rootNode.nodes[rootNodesLength - 1].whenSome;
+                        } else {
+                            rootNode.nodes[rootNodesLength - 1].nodes = rootNode.nodes[rootNodesLength - 1].whenNone;
+                        }
+                    } else {
+                        rootNode.nodes[rootNodesLength - 1].nodes = [
+                            ...rootNode.nodes[rootNodesLength - 1].nodes,
+                            constructedNode,
+                        ];
+                    }
```
algomaster99, line 311: This and some chunks below it more or less look similar. You increase the level of nesting for each of them on observation. Maybe create a generic function for them which takes in `rootNode` and `level` and recurses `level` number of times to return the desired nodes.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -32,8 +32,11 @@ class ToCiceroMarkVisitor {
         // All the nodes generated from given OOXML
         this.nodes = [];
 
-        // contains the realtionship part of a given OOXML
+        // contains the relationship part of a given OOXML
         this.relationshipXML = [];
+
+        // contains the nodes present in a conditional or optional OOXML
+        this.conditionalOrOptionalNodes = [];
```
algomaster99, line 39: Same logic as below. I think it would be easier to implement it here because you actually return the nodes here. So you can recursively create an entire node for Optional node without having this array as a side-effect.

```
constructCiceroNodes() {
   if OPTIONAL { 
      ooxmlToBeDisplayed = constructCiceroNodes(whenSome) // ofcouse this depends upon `hasSome` so you would need a check for it.
      ooxmlToBeHidden = constructCiceroNodes(whenNode)
  }
}
```

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -334,14 +442,41 @@ class ToCiceroMarkVisitor {
                         if (runTimeProperties.attributes['w:fill'] === 'F9F2F4') {
                             shadeCodePresent = true;
                         }
+                    } else if (runTimeProperties.name === 'w:vanish') {
+                        vanishPropertyPresent = true;
+                    } else if (runTimeProperties.name === 'w:rFonts') {
+                        if (runTimeProperties.attributes['w:ascii'] === 'Baskerville Old Face') {
```
algomaster99, line 448: Why is there a different font family for Optional?

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -334,14 +442,41 @@ class ToCiceroMarkVisitor {
...
+                    } else {
+                        nodeInformation.whenSomeType = true;
+                        nodeInformation.hasSome = true;
+                    }
+                }
             } else if (runTimeNodes.name === 'w:t') {
                 if (calledBy === TRANSFORMED_NODES.codeBlock) {
                     ooxmlTagTextValue += runTimeNodes.elements ? runTimeNodes.elements[0].text : '';
+                } else if (calledBy === TRANSFORMED_NODES.optional) {
+                    ooxmlTagTextValue = runTimeNodes.elements ? runTimeNodes.elements[0].text : ' ';
```
algomaster99, line 477: If runTimeNodes.element is empty, why do we need a `' '` (space)?

algomaster99 (changes requested):
Dropped some comments which I think you really need to look into. Please try to incorporate them. I will try to take out some time and refactor your code a bit over the weekend.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
K-Kumar-01: @algomaster99 
simply using `w:vanish` we cant tell whether this is of `whenSome` type or `whenNone`. So, if 
```
vanish+font=whenNone
vanish=whenSome
```

As for the different font family, I did not want to make it similar to the existing ones or the one used in `code` or `codeblock`.

packages/markdown-docx/src/ToOOXMLVisitor/rules.js
K-Kumar-01: @algomaster99 
I have already linked it.

algomaster99: I cannot see the link. I meant the link to the OOXML docs. Otherwise, you can explain that it is equivalent to `display: none`.

K-Kumar-01: @algomaster99 Oh sure. I meant you were asking for it to be present in the docstrings. I will add a link.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
algomaster99: I am not sure if it's true and it's merely based on observation. I have noticed that `whenNone` is always an empty array in case of `Optional`.

K-Kumar-01: @algomaster99 
It is not true. It can contain some nodes as visible [here](https://docs.accordproject.org/docs/markup-templatemark.html#optional-blocks). When this is pasted in https://templatemark-dingus.netlify.app/, whenNone can have some nodes.

packages/markdown-docx/src/constants.js
K-Kumar-01: @algomaster99 
This separator can be used.

K-Kumar-01:
@algomaster99 
Made the desired changes.

K-Kumar-01:
@algomaster99 @dselman 
Requesting your reviews.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -69,9 +77,10 @@ class ToOOXMLVisitor {
      * @param {string}  value           Text value of the node
      * @param {Array}   nodeProperties  Properties of the node
      * @param {boolean} calledByCode    Is function called by code node or not
+     * @param {object}  conditions      else and false conditions of optional and conditional nodes
      * @returns {string} Generated OOXML
      */
-    generateTextOrCodeOOXML(value, nodeProperties, calledByCode = false) {
+    generateTextOrCodeOOXML(value, nodeProperties, calledByCode = false, conditions) {
```
algomaster99, line 83: Instead of introducing another `conditions attribute, we should combine its attributes, `isTraversingFalseCondition` and `isTraversingElseCondition`, with `nodeProperties` because their behaviour and function are essentially the same. You can probably create your own properties to distiguish between the above two attributes, whose difference is not clear to me.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -266,7 +336,10 @@ class ToOOXMLVisitor {
      * @returns {string} OOXML string
      */
     toOOXML(ciceromark) {
-        this.traverseNodes(ciceromark, []);
+        this.traverseNodes(ciceromark, [], TRANSFORMED_NODES.document, {
+            isTraversingElseCondition: false,
+            isTraversingFalseCondition: false,
```
algomaster99, line 341: I will try to explain why this is a bad solution. `toOOXML` is supposed to deal with the _entire_ ciceromark which contains multiple type of nodes. The properties passed here must be true for all the nodes inside it. However, when you pass such conditions which are only true for the optional node, it does not respect the previously stated rule. Moroever, it is highly possible that these properties are not used because the CiceroMark doesn't have an `Optional` node.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
```diff
@@ -240,6 +281,34 @@ class ToOOXMLVisitor {
...
                             this.tags = [];
+                        } else if (this.getClass(subNode) === TRANSFORMED_NODES.optional) {
+                            conditions.isTraversingFalseCondition = !subNode.hasSome;
+                            // traverse the whenSome properties now
+                            const whenSomeOOXML = this.traverseNodes(
+                                subNode.whenSome,
+                                properties,
+                                TRANSFORMED_NODES.optional,
+                                conditions
+                            );
```
algomaster99, line 292: Well done! This is so clear and concise.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -334,14 +364,40 @@ class ToCiceroMarkVisitor {
...
+                    }
+                } else {
+                    if (fontFamilyPresent) {
+                        nodeInformation.whenSomeType = false;
+                        nodeInformation.hasSome = false;
+                    } else {
+                        nodeInformation.whenSomeType = true;
+                        nodeInformation.hasSome = true;
+                    }
+                }
```
algomaster99, line 394: Need to understand this before I can give reviews here.

If `hasSome` is `true`, then `whenSome` is rendered and if `hasSome` is false, `whenNone` is rendered. So there are only two possible conditions which should be encompassed by a **single** boolean variable. Why are there two - `nodeInformation.hasSome` and `nodeInformation.whenSomeType`?

The same logic applies to `toOOXML` because you have used two boolean variables there too, `isTraversingFalseCondition` and `isTraversingElseCondition`.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -392,7 +463,7 @@ class ToCiceroMarkVisitor {
                     let text = '';
                     for (const codeBlockSubNode of subNode.elements) {
                         if (codeBlockSubNode.name === 'w:r') {
-                            text = this.fetchFormattingProperties(codeBlockSubNode, TRANSFORMED_NODES.codeBlock);
+                            text = this.fetchFormattingProperties(codeBlockSubNode, TRANSFORMED_NODES.codeBlock, {});
```
algomaster99, line 466: `nodeInformation` was `null` before. How does an empty object help?

packages/markdown-docx/src/ToCiceroMarkVisitor.js
```diff
@@ -305,17 +332,20 @@ class ToCiceroMarkVisitor {
      * @param {Array}   node              Node to be traversed
      * @param {string}  calledBy          Parent node class that called the function
      * @param {object}  nodeInformation   Information for the current node
-     * @returns {string} Value in <w:t> tags
+     * @returns {string} NodeInformation array if optional else Value in <w:t> tags
```
algomaster99, line 335: What exactly are the contents of the array if node is optional? You can encode that with `{string}` by using a union operator.

algomaster99 (changes requested):
Left some comments. There are some things I don't understand so I might give more reviews on this PR in future.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
K-Kumar-01: @algomaster99 
if `hasSome` is true, then what nodes are `whenSome` type. there is no clear way to distinguish that. That's why we use font properties for determining whether its `whenSome` or `whenNone` type.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
K-Kumar-01: @algomaster99 
There were errors for codeblock test case, so I passed an empty object.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
I think that we can pass an empty object here. It will work and there won't be any need for explicitly declaring these properties at the time of calling.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
K-Kumar-01: @algomaster99 
I think that conditions which I was thinking of renaming to `parentProps` will be better suited in this case. The reason being conditions applies to a set of nodes and subnodes. `Nodeinformation` is specific to a particular node. Also, it would make the nodeInformation complex.

In the list transformer we also need certain properties of parent, which will obviously increase the `nodeProperties` as we go down.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
algomaster99: I would suggest looking for the reason for the error instead of introducing such workarounds.

packages/markdown-docx/src/ToOOXMLVisitor/index.js
algomaster99: > I think that we can pass an empty object here.

You completely ignored "The properties passed here must be true for all the nodes inside it. ". Things should be abstracted as much as possible so that developers don't have to deal with such minute stuff while working on a higher level.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
K-Kumar-01: @algomaster99 i know the reason and will explain in the meeting.

packages/markdown-docx/src/ToCiceroMarkVisitor.js
algomaster99: > if hasSome is true, then what nodes are whenSome type

```json
{
          "$class": "org.accordproject.ciceromark.Optional",
          "hasSome": false,
          "whenSome": [
            {
              "$class": "org.accordproject.commonmark.Text",
              "text": " except for Force Majeure cases in a "
            },
            {
              "$class": "org.accordproject.ciceromark.Variable",
              "value": "100.0",
              "name": "miles",
              "elementType": "Double"
            },
            {
              "$class": "org.accordproject.commonmark.Text",
              "text": " miles radius,"
            }
          ],
          "whenNone": [
            {
              "$class": "org.accordproject.commonmark.Text",
              "text": "This applies even in case a force majeure."
            }
          ],
          "name": "forceMajeure",
          "elementType": "org.accordproject.latedeliveryandpenalty.Distance",
          "nodes": [
            {
              "$class": "org.accordproject.commonmark.Text",
              "text": "This applies even in case a force majeure."
            }
          ]
        }
```
Taking your testcase as example, I think `Optional` node follows an algorithm.

1. If `hasSome` is true, `whenSome` is rendered.
2. If `hasSome` is false, `whenNone` is rendered.

@dselman could you please confirm this? Or else, please link the file where the model of `Optional` is defined.

K-Kumar-01:
@algomaster99 
Made the changes. Let me know if there are any other changes required.

packages/markdown-docx/src/ToOOXMLVisitor/rules.js
```diff
@@ -283,7 +283,8 @@ const OPTIONAL_RULE = (title, tag, value, type) => {
 const VANISH_PROPERTY_RULE = () => '<w:vanish/>';
 
 /**
- * Inserts Baskerville Old Face font family OOXML tag.
+ * Inserts Baskerville Old Face font family OOXML tag to distinguish nodes of whenNone/whenFalse
```
algomaster99, line 286: Try to put yourself in someone who is not that familiar with this code.

They try to interpret this function and they found out that it has something to do with optional or condition nodes. They look at the CiceroMark properties and see that in fact, `whenSome` and `whenNone` exist. However, they would definitely question "what are `whenTrue` and `whenFalse` for?" Given then context, they would deduce that `whenTrue` and `whenFalse` are also attributes in the CiceroMark which is false!

In fact this is true wherever you have metioned the above terms so you need to rectify that. Probably a better wording would be:

```
Inserts a different fontface so that the hidden contents can be distinguished. There can be two types of hidden content -
1) when `hasSome` is true, the content inside `whenNone` is hidden.
2) when `hasSome` is false, the content inside `whenSome` is hidden.
```

I know this takes time and it may not be commensurate for the time you spend thinking over it. But I can promise you that you only will forget the same code you have written 4 months later given you don't work on this project for that long.

K-Kumar-01: @algomaster99 Actually, this is not quite the right description. the font family is for `else` condition always. It does not matter whether they are hidden or visible.

algomaster99: If they are visible, then we do not want to change the font.

K-Kumar-01: @algomaster99 
Does this sound right?
```md
 Inserts a different font family so that the conditional/optional content can be distinguished. They can be of following types:-
 1) `whenNone` nodes for optional(to be distinguished from `whenSome`)
 2) `whenFalse` nodes for conditional(to be distinguished from `whenTrue`)
```

algomaster99: Okay, so the font face has been introduced to handle conditional as well? I thought you would make a separate PR for that.

K-Kumar-01: @algomaster99 
It will automatically deal with that when I make the PR for conditional. `Optional` and `Conditional` are identically same. Just some node keys changing.

K-Kumar-01: @algomaster99 
I will update the JSDOC for conditional in that PR. Currenly, I will only make it for optional. Does it sounds fine?

algomaster99: Sounds good. You could have applied the same logic for the variables too but that's okay. We can merge this one.

Anyway, just try to focus on one feature at a time. It causes confusion otherwise.

algomaster99:
@K-Kumar-01 Shall I proceed for merge if all tests are passing?

K-Kumar-01:
@algomaster99 Sure
was pushing the updated JSDOC comment for optional for now.
WIll later change it for both conditional and optional as mentioned.

PR #474, opened 2022-01-29
fix(DateTime/draft): optional DateTime variable default value set to undefined -cicero#687 

OmkarAcharekar:
Signed-off-by: Omkar Acharekar <omkaracharekar12@gmail.com>

# Closes #<accordproject/cicero#687>

### Changes

- <ONE>Added condition to handle the undefined value that get generated when optional dateTime variable is legitimately omitted.

### Author Checklist
- [x] Merging to `master` from `fork:branchname`


packages/markdown-template/lib/plugins/DateTime/draft.js
```diff
@@ -25,8 +25,14 @@ dayjs.extend(utc);
  * @returns {string} the text
  */
 function dateTimeDrafter(value,format) {
```
dselman, line 27: I think this will also have to roundtrip - I.e. the string "undefined" should set the DateTime to null. Which makes me think that this is not a `DateTime` specific issue. @jeromesimeon what do you think?

dselman (changes requested):

packages/markdown-template/lib/plugins/DateTime/draft.js
OmkarAcharekar: @dselman can you suggest any changes?

martinhalford: @jeromesimeon @dselman - To me, this fix looks like it might work. However, I'm not sure about the roundtrip issue  that Dan mentions.  

I am noticing that `null` DateTime variables are defaulting to `today` in other parts of the code - which is annoying.  

What's the best way forward?  At the moment we can't use `optional DateTime` variables in non-US grammar, which is very limiting.

Another option, can we maybe make DateTime variables default to a non-ambiguous universal format throughout the codebase?  (e.g. DD-MMMM-YYYY) A potentially breaking change, I'm assuming?

martinhalford (commented):
Comments inline

PR #483, opened 2022-02-21
fix(markdown-pdf) - Allow pdfMakeToPdfStreamWithCallback method of PdfTransformerBase's class to override default style

martinhalford:
Signed-off-by: Shivam upadhyay <shivam.upadhyay@codinova.com>

Allow pdfMakeToPdfStreamWithCallback method of PdfTransformerBase's class to override default style

### Changes
- <ONE> Change pdfMakeToPdfStreamWithCallback to include default style only if provided input does not contains defaultStyle

### Screenshots or Video
Screenshot 2022-02-21 at 4 46 00 PM

### DIFF
--- a/packages/markdown-pdf/src/PdfTransformerBase.js
+++ b/packages/markdown-pdf/src/PdfTransformerBase.js
@@ -103,12 +103,15 @@ class PdfTransformerBase {
      */
     static pdfMakeToPdfStreamWithCallback(progressCallback) {
         return (input, outputStream) => {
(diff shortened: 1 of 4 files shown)


dselman (approved):

jeromesimeon (changes requested):
looks good! Two requests:
1. it's only one change to one file, could we please just have one commit instead of 7? this could help also handle the following:
2. DCO please!

martinhalford:
Closing PR in order to regenerate without spurious commits.

PR #637, opened 2025-12-25
Fix HTML rendering of relationship variables in CiceroMark

Aradhana2803:
# Fix HTML rendering of relationship variables in CiceroMark

Related to accordproject/template-playground#13 

### Summary
Relationship-typed CiceroMark variables were rendered in HTML as quoted
`resource:` strings instead of a readable identifier. This PR fixes the
HTML rendering logic so relationship variables display correctly.[TemplateMark not rendering relationships "resource" data  #13]

### Changes
- Update `ToHtmlStringVisitor` to detect relationship variables via `elementType`
- Render a clean relationship identifier instead of the raw serialized value
- Preserve existing rendering behavior for primitive variable types

### Flags
- No breaking changes
- Change is limited to HTML rendering only (CiceroMark JSON unchanged)

### Screenshots or Video
Not applicable — behavior verified via existing HtmlTransformer test suite.

### Author Checklist
- [x] Ensure you provide a DCO sign-off for your commits using the `--signoff` option of git commit
- [x] Commits messages follow AP format
- [x] Extend the documentation, if necessary
- [x] Merging to `main` from `fork:branchname`

### DIFF
--- a/packages/markdown-html/src/HtmlTransformer.test.js
+++ b/packages/markdown-html/src/HtmlTransformer.test.js
@@ -121,3 +121,63 @@ describe('ciceromark <-> html', () => {
         });
     });
 });
+
+describe('renderVariableValue - relationship variables', () => {
+    it('extracts identifier from a quoted resource URI value', () => {
+        const ciceroMarkJson = {
+            '$class': 'org.accordproject.commonmark@0.5.0.Document',
+            'xmlns': 'http://commonmark.org/xml/1.0',
+            'nodes': [{
+                '$class': 'org.accordproject.commonmark@0.5.0.Paragraph',
+                'nodes': [{
+                    '$class': 'org.accordproject.ciceromark@0.6.0.Variable',
+                    'value': '"resource:org.accordproject.organization@0.2.0.Organization#Party A"',
+                    'identifiedBy': 'identifier',
+                    'name': 'buyer',
+                    'elementType': 'org.accordproject.organization@0.2.0.Organization'
+                }]
+            }]
+        };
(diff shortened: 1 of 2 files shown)


Aradhana2803:
This PR fixes the issue reported in template-playground#13 where relationship-typed CiceroMark variables were rendered as quoted resource: strings instead of a readable identifier. HTML rendering now preserves the identifier cleanly, while keeping other variable types unchanged.Verified that `Party` relationship variables now render clean identifiers in HTML
without quotes. All existing HtmlTransformer tests pass.

sanketshevkar (changes requested):
Can you please add unit tests covering this case?

mttrbrts:
@sanketshevkar I've pushed an updated commit on top of this branch that addresses your review.

**`renderVariableValue` — more robust relationship detection**

The original check used `thing.elementType.startsWith('org.accordproject.party@')`, which only handled one hardcoded namespace. I've replaced it with a check on `thing.identifiedBy`, which is the canonical Concerto marker for relationship/reference types regardless of namespace. The logic now only rewrites the display value when it's a full resource URI (`resource:namespace#identifier`), extracting the identifier after `#`. Plain quoted values like `"Party A"` are passed through unchanged, so existing roundtrip tests are unaffected.

**Unit tests added**

Three cases in a new `describe` block in `HtmlTransformer.test.js`:
1. Quoted resource URI value → renders as clean identifier, no `resource:` prefix
2. Bare (unquoted) resource URI value → same
3. Primitive variable (no `identifiedBy`) → value unchanged

All 65 existing tests still pass.

PR #652, opened 2026-04-07
fix(markdown-common): retain empty nodes arrays in AST to satisfy strict=true in concerto-v4

### LINKED ISSUE #651: [Bug] Concerto v4 Compatibility: Missing 'nodes' array in AST causes ValidationException
# Bug Report 🐛

@mttrbrts , @dselman As asked in the last workgroup meeting I am opening this issue describing the errors I came across while testing concerto-v4 on markdown-transform. 

While I was  testing `template-engine` against Concerto v4, I encountered consistent runtime crashes. The root cause traces back to the ASTs generated by `markdown-transform`. 

Concerto v4 introduces strict array enforcement. If a property is defined in the metamodel as an array (e.g., the `nodes` property in CommonMark/TemplateMark elements), v4 strictly demands that the property exists and is an array, even if it is empty (`nodes: []`). 

Currently, `markdown-transform` is "lazy" in its AST generation—if a markdown element has no children, it either omits the `nodes` property entirely or leaves it as `undefined`. When this AST is passed up the stack, the v4 engine's strict runtime validation rejects it and crashes the execution.

### Error Message
```text
ValidationException: Expected property 'nodes' to be an array
```

### Steps to Reproduce
1. Initialize a Concerto v4 engine with the standard CommonMark/TemplateMark metamodels.
2. Use `markdown-transform` to parse a valid markdown string that results in leaf nodes or blocks without child elements.
3. Pass the resulting JSON AST into the Concerto v4 validator.
4. The v4 engine will throw a `ValidationException` at runtime due to the missing `nodes` arrays.

### Impact
This is currently a hard blocker for upgrading `template-engine`, `cicero`, and downstream applications (like APAP) to Concerto v4.

### Proposed Solutions for Discussion
We have a few potential paths to resolve this, and I'd love your architectural guidance on which direction to take:
1. **Parser Update:** Update the core parsing logic within `markdown-transform` to guarantee `nodes: []` is always instantiated for elements that require it.
2. **Recursive AST Patch (Post-Processing):** Introduce a sanitization utility that recursively walks the AST immediately after parsing, forcefully injecting `nodes: []` wherever the metamodel demands it (we have a local script successfully doing this to bypass the errors).
3. **Metamodel Relaxation:** Update the underlying Concerto metamodel for CommonMark to make the `nodes` array optional, though this might conflict with the design philosophy of v4's strictness.

Let me know which approach aligns best with the v4 roadmap, and I'd be happy to spearhead the PR for it!

rockaxorb13:
### Description
fixes #651 
This PR fixes a critical AST validation crash that occurs when using  Concerto v4 engine (specifically `4.0.0-beta.1` and later) alongside `markdown-transform`. I uncovered this bug 2-3 weeks ago and and the fix was easier than I anticipated!

**The Root Cause:**
As described in the issue:
In `packages/markdown-common/lib/FromMarkdownIt.js` (`blockToCommonMark`), the parser was explicitly deleting the `nodes` array if a markdown element (like an empty clause, list, or blockquote) had no children. While Concerto v3 permitted this "lazy AST" where the array was simply `undefined`, the new strict type-checking in Concerto v4 throws a fatal `ValidationException: Expected property 'nodes' to be an array` because `nodes` is defined as a mandatory array in the metamodel.

**The Fix:**
Removed the aggressive deletion block (`delete node.nodes`) in `FromMarkdownIt.js` so that empty elements correctly retain an empty array (`"nodes": []`). 

### Testing / Validation
- Tested against Concerto v3: Fully backward compatible. Ran the full test suite and updated Jest snapshots successfully.
- Tested against Concerto v4: Prevents the `ValidationException` crash in strict parsing environments.

**Standalone Script Output Proof:**
I made a script to directly test this:
By feeding an empty clause (`{{#clause testClause}}\n{{/clause}}`) into `CiceroMarkTransformer`, we can see the exact structural change to the AST:

**Before (Fails v4 Validation):**
```json
--- RAW AST OUTPUT ---
{
  "$class": "org.accordproject.ciceromark@0.6.0.Clause",
  "name": "testClause"
}
❌ BUG REPRODUCED: 'nodes' array was deleted!
```
**After (Passes v4 Validation):**
```json
--- RAW AST OUTPUT ---
{
  "$class": "org.accordproject.ciceromark@0.6.0.Clause",
  "name": "testClause",
  "nodes": []
}
✅ FIX WORKS: 'nodes' array is present!
```

### DIFF
--- a/packages/markdown-cicero/lib/__snapshots__/CommonMarkSpec.test.js.snap
+++ b/packages/markdown-cicero/lib/__snapshots__/CommonMarkSpec.test.js.snap
@@ -484,14 +484,17 @@ exports[`markdown-spec converts ATX headings-49 to concerto 1`] = `
     {
       "$class": "org.accordproject.commonmark@0.5.0.Heading",
       "level": "2",
+      "nodes": [],
     },
     {
       "$class": "org.accordproject.commonmark@0.5.0.Heading",
       "level": "1",
+      "nodes": [],
     },
     {
       "$class": "org.accordproject.commonmark@0.5.0.Heading",
       "level": "3",
+      "nodes": [],
     },
   ],
   "xmlns": "http://commonmark.org/xml/1.0",
@@ -1656,6 +1659,7 @@ exports[`markdown-spec converts Block quotes-209 to concerto 1`] = `
   "nodes": [
     {
       "$class": "org.accordproject.commonmark@0.5.0.BlockQuote",
+      "nodes": [],
     },
   ],
   "xmlns": "http://commonmark.org/xml/1.0",
@@ -1668,6 +1672,7 @@ exports[`markdown-spec converts Block quotes-210 to concerto 1`] = `
   "nodes": [
     {
       "$class": "org.accordproject.commonmark@0.5.0.BlockQuote",
+      "nodes": [],
     },
   ],
   "xmlns": "http://commonmark.org/xml/1.0",
@@ -10394,6 +10399,7 @@ exports[`markdown-spec converts Link reference definitions-187 to concerto 1`] =
     },
     {
       "$class": "org.accordproject.commonmark@0.5.0.BlockQuote",
+      "nodes": [],
     },
   ],
   "xmlns": "http://commonmark.org/xml/1.0",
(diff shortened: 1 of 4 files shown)


mttrbrts (approved):
Great find!

